PR #21, opened 2021-03-12
Bug/ios fixes

DavidClawson:
I changed the Podfile's build target to iOS 10
Also, added the aries-framework-javascript.tgz to the repo, and modified the .gitignore so it'd stay there.


JamesKEbert (approved):
LGTM! Thanks!

PR #23, opened 2021-03-12
Installed react-native-config to support the use of the .env URLs

DavidClawson:
Added react-native-config to support URLs in .env file.


JamesKEbert (approved):
LGTM

PR #61, opened 2021-07-07
feat: add @aries-framework/redux-store

TimoGlastra:
Replacement of #57 because of DCO issues

the nesting of a package inside a package is giving a lot of issues. We'll move this over to AFJ ASAP when the monorepo is ready (real soon)


karimStekelenburg:
Do DOCs behave differently in mono-repos or is it purely a problem with nesting? Since AFJ is also a Hyperledger repo we might end up with the same trouble..

TimoGlastra:
DCO is unrelated to this. I'm not sure why the DCO issues arose, but it's just a modification to the commit message.

karimStekelenburg (dismissed):

karimStekelenburg (approved):

PR #65, opened 2021-07-19
Feature/u iupdate

manghuli:
# Summary of Changes

complete UI overhaul, start to a component library, and a general UI pattern to be followed going forward. react-router-native was completely replaced by react-navigation to give as native an experience as possible. Global styles for main colors and generic styles. 

# Related Issues

# Pull Request Checklist


TimoGlastra (commented):
Awesome!

I see you removed the Indy.Framework. Any specific reasoning behind this?

manghuli:
> Awesome!
> 
> I see you removed the Indy.Framework. Any specific reasoning behind this?

oh, that was an accident. Getting that fixed now, thank you

App/components/listItems/ContactListItem.tsx
```diff
@@ -0,0 +1,38 @@
...
+}
+
+const ContactListItem: React.FC<Props> = ({ contact }) => {
+  const date = JSON.stringify(contact.createdAt)
+
+  return (
+    <View key={contact.contact_id} style={styles.container}>
+      <Title>{contact.alias || contact.invitation.label}</Title>
+      <Text>{contact.did}</Text>
+      <Text style={styles.date}>{`${Number(date.slice(6, 8))}/${Number(date.slice(9, 11))}/${date.slice(1, 5)}`}</Text>
```
JamesKEbert: Would like to make this less fragile, now or in the future

App/components/listItems/NotificationListItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const NotificationListItem: React.FC<Props> = ({ notification }) => {
+  const navigation = useNavigation()
+
+  const decideTitle = () => {
+    switch(notification.type) {
+      case 'CredentialRecord':
+        return 'Credential Offer'
+      default:
+        return ''
```
JamesKEbert: Maybe we return the notification type and throw a warning?

App/components/views/AppHeader.tsx
```diff
@@ -0,0 +1,52 @@
...
+import AppStyles from '../../../assets/styles'
+
+interface AppHeaderProps {
+  headerText?: string
+}
+
+function AppHeader(props: AppHeaderProps) {
+  return (
+    <View style={styles.header}>
+      <TouchableOpacity onPress={() => {}}>
```
JamesKEbert: Unnecessary TouchableOpacity?

App/navigators/AuthenticateStack.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+interface Props {
+  setAuthenticated: () => void
+}
+
+const AuthenticateStack: React.FC<Props> = ({ setAuthenticated }) => {
+  const [firstLogin, setFirstLogin] = useState(true)
+
+  const checkFirstLogin = async () => {
+    // await Keychain.resetGenericPassword({ service: 'passcode' })
```
JamesKEbert: Are we still checking if this is a new install? We have been using `AsyncStorage` for this.

App/screens/Home.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const agentContext = useContext<any>(AgentContext)
+
+  const [notifications, setNotifications] = useState<any>([])
+
+  const getNotifications = async () => {
+    const credentials = await agentContext.agent.credentials.getAll()
+    const credentialsToDisplay: any = []
+
+    credentials.forEach((c: any) => {
+      if (c.state !== CredentialState.Done) {
```
JamesKEbert: Since I think we only want to display a notification for a credential that is in a state Offered (at least for now)? Or am I missing something here?
```suggestion
      if (c.state === CredentialState.OfferReceived) {
```

App/screens/PinCreate.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const confirmEntry = (x: any, y: any) => {
+    if (x.length < 6 || y.length < 6) {
+      Alert.alert('Pin must be 6 digits in length')
+    } else if (x !== y) {
+      textInput.clear()
+      secondTextInput.clear()
+      Alert.alert('Pins entered do not match')
+    } else {
+      passcodeCreate(x)
+      props.navigation.navigate('Enter Pin')
```
JamesKEbert: I think the user probably doesn't need to reenter their pin after they created it? So we should probably authenticate instead?

JamesKEbert (changes requested):
Overall this is great! Love the code restructuring and UI alterations. There are a couple of portions of the credential issuance flow that we likely want to alter, but we can do that as a followup PR. 
I would like to see if we can address all the warnings thrown though. I've captured the following two (I don't recall if there's more or not).

> VirtualizedLists should never be nested inside plain ScrollViews with the same orientation because it can break windowing and other functionality - use another VirtualizedList-backed container instead.

> Warning: Each child in a list should have a unique "key" prop.

I've also left a couple of requests/questions, but again, great work!

App/components/listItems/ContactListItem.tsx
manghuli: Want me to add that date parsing library thing we added to sita and use that?

App/navigators/AuthenticateStack.tsx
manghuli: I'm gonna need an explanation on how you guys decided to do this

App/components/listItems/ContactListItem.tsx
JamesKEbert: Yeah I think that'd be a decent approach here--my biggest concern here was that if the date format changes at all the slicing will definitely blow up in some ways.

manghuli: ya, it was supposed to be a stopgap for suuureee

App/components/listItems/NotificationListItem.tsx
```diff
@@ -15,21 +15,12 @@ interface Props {
...
   return (
     <TouchableOpacity
       style={styles.container}
       onPress={() => navigation.navigate('Credential Offer', { notification })}
     >
-      <Text style={styles.title}>{decideTitle()}</Text>
+      <Text style={styles.title}>{notification.type || 'Notification'}</Text>
       <Icon name="chevron-right" color={textColor} size={30} />
     </TouchableOpacity>
   )
```
JamesKEbert: This would likely need subsequent updates to account for different notification types, right?

manghuli: na, just whatever type is on the object from the system will be rendered, This will change with the cred flow updates though

JamesKEbert: Yes, and when you tap on the notification right now it's set to take you to the credential offer screen, even it's not an offer I think

manghuli: oh! correct, we'll have to make a more generic notificationDetails screen or something like that

JamesKEbert (approved):
This looks good to me. There's some pieces I'd still like to iterate on style-wise and pieces of the credential issuance flow, but I think this is overall headed in a good direction.

karimStekelenburg (approved):
Looks good to me :)

Issue #92, opened 2021-10-20
The workflows are not able to run to completion


jleach:
The GitHub Actions (workflows) are not able to run because there are too many LINTing issues.

PR #93, opened 2021-10-20
Add workflow to build for iOS and Android

jleach:
# Summary of Changes

Added a GitHub Actions based wrokflow to make sure the application combines for both iOS and Android.

# Related Issues

I thought we could address #92 with two PRs. The first hist to beef up the workflows by demonstrating that the project can compile for both iOS and Android. This will help ensure no breaking changes are included by contributors. The down side, is that it takes about 25 mins for the iOS build to run on GitHub hardware. Not an overly big issue since its all background work.

The second pull request would be to fix all the LINT issues and then add `npm run lint` back into this workflow or another just for listing. It may also be worth adding husky support to run a lint check on commit to not allow un-formatted code to be pushed out in the first place.

**NOTE**: This will burn through allot of build minutes for private repositories. Build minutes on `macOS` are billed at 10x so a 25 min run-time will be 250 minutes. 

# Pull Request Checklist

- [X] Updated **documentation** for changed code and new or modified features.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,10 +1,11 @@
 name: Continuous Integration
 
 on:
-  pull_request:
-    branches: [main]
-  push:
-    branches: [main]
+  workflow_dispatch:
+  # pull_request:
+  #   branches: [main]
+  # push:
+  #   branches: [main]
 
 jobs:
   integration-test:
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,0 +1,155 @@
+name: Native Build & Test
+
+env:
+  cacheId: '3' # increment to expire the cache
+
+on:
+  push:
+    paths:
+      - 'ios/**'
+      - 'android/**'
+      - '**/package-lock.json'
+
+jobs:
+  build-ios:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Configure node
+        uses: actions/setup-node@v1
+        with:
+          node-version: "12.x"
+          registry-url: "https://registry.npmjs.org"
+      
+      - name: Configure ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 2.6
+
+        # GitHub recommends not caching node_modules but rather
+        # .npm because it can break across Node versions and
+        #  won't work with npm ci.
+      - name: Cache node modules
+        uses: actions/cache@v1
+        id: npm-cache
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ env.cacheId }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-${{ env.cacheId }}-
+
+      - name: Cache pod dependencies
+        id: pod-cache
+        uses: actions/cache@v1
+        with:
+          path: ios/Pods
+          key: ${{ runner.os }}-pods-${{ env.cacheId }}-${{ hashFiles('**/Podfile.lock ') }}
+          restore-keys: |
+            ${{ runner.os }}-pods-${{ env.cacheId }}-
+
+      # Watch for changes to the `App` and `iOS` paths, use
+      # git for cache keys.
+      - name: Generate cache key
+        run: |
+          echo $(git rev-parse HEAD:App) > ./dd-cache-key.txt
+          echo $(git rev-parse HEAD:ios) >> ./dd-cache-key.txt
+
+      - name: Cache derived data
+        uses: actions/cache@v1
+        with:
+          path: ios/xbuild/Build
+          key: ${{ runner.os }}-dd-xcode-${{ env.cacheId }}-${{ hashFiles('**/dd-cache-key.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-dd-xcode-${{ env.cacheId }}-
+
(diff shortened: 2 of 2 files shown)


jleach:
@JamesKEbert Your thoughts on adding this to the main repo. Because RN depends on 3rd-party components integrated directly into the iOS and Android projects that require linking or `pod install` it would be good to make sure that no contributions break-the-build. The down side, as noted above it takes some time for the iOS build to run on GitHub provided hardware. The up side is that it is free and a background task. Could be moved to run on merges to master only but by that times breaking changes are harder to unwind.

TimoGlastra:
I like it! Would be good to eventually add Detox testing to make sure the basic flow of creating connections, receiving credentials and sharing proofs is still working as expected.

Do you think caching could help with the iOS build times. Incremental builds are way faster than cold builds

jleach:
@TimoGlastra I've put in as much caching as I can. Caching build artifacts seems like it would help (Android only takes 5min) but I haven't figured out how to do it (despite some effort). I can take one last try but if I can't speed it up I think there is still value in a binary compile even if its slow.

EDIT: I don't think additional caching will help. Even on my MacBook Pro the build time is exceedingly long compared to Android. Judging by search results its not uncommon to react native projects.

JamesKEbert:
@jleach I would fully be in favor of adding this to the main repo! 

> It may also be worth adding husky support to run a lint check on commit to not allow un-formatted code to be pushed out in the first place.

We should be using Husky right now IIRC for ensuring DCO signoff, so it wouldn't be that hard to add a linting check too.

jleach:
No progress on the build time for iOS. It's not using the cache for some reason. I think it may be `npm ci` or `pod install` causing the intermediate build artifacts to expire. I've updated the path triggers to only run the native build when something meaningful changes. I don't think it needs to run when a `*.tsx,*.ts` file change as that would trigger a run-time error. Only changes to `ios/**`, `android/**` or `package-lock.json` will trigger this workflow.

jleach:
Added this note to my original description above:
```
NOTE: This will burn through allot of build minutes for private repositories. Build
minutes on macOS are billed at 10x so a 25 min run-time will be 250 minutes.
```

.github/workflows/continuous-integration.yml
```diff
@@ -1,10 +1,11 @@
...
 on:
-  pull_request:
-    branches: [main]
-  push:
-    branches: [main]
+  workflow_dispatch:
+  # pull_request:
+  #   branches: [main]
+  # push:
+  #   branches: [main]
```
TimoGlastra, line 8: I'd like to keep running this on PRs and pushes to main

.github/workflows/main.yaml
```diff
@@ -0,0 +1,155 @@
...
+
+env:
+  cacheId: '3' # increment to expire the cache
+
+on:
+  push:
+    paths:
+      - 'ios/**'
+      - 'android/**'
+      - '**/package-lock.json'
```
TimoGlastra: To not run it on every branch and push I normally do it on pull_request and push to main. This way it will always run before a PR is merged (and we can set the job to required for the PR to be mergeable)

```
on:
  pull_request:
    branches: [main]
  push:
    branches: [main]
```

TimoGlastra (changes requested):
This is great @jleach!

Only the comment should be fixed I think, but for the rest good to go

.github/workflows/continuous-integration.yml
jleach: I agree. I disabled it for now until all the LINT issues are resolved. I'm going to clean them up and re-enable this workflow as part of a new PR I'll start on today.

.github/workflows/main.yaml
jleach: Are you thinking something like this? There may be some value removing `paths:` from pushes to main just to be sure `main` always compiles.

```
on:
  pull_request:
    branches: [main]
    paths:
      - 'ios/**'
      - 'android/**'
      - '**/package-lock.json'
  push:
    branches: [main]
    paths:
      - 'ios/**'
      - 'android/**'
      - '**/package-lock.json'
 ```

Comment (TimoGlastra) [reply .github/workflows/main.yaml]: Sorry forgot to comment on your reply. But yes, that what what I was thinking about. I think you suggestion to remove paths for pushes to main makes sense

File: .github/workflows/main.yaml [hunk 3]
```diff
@@ -0,0 +1,155 @@
+name: Native Build & Test
+
+env:
+  cacheId: '3' # increment to expire the cache
```
TimoGlastra, line 4: Is there a specific reason this is needed?

TimoGlastra (dismissed):
This is great @jleach!

.github/workflows/main.yaml
jleach: @TimoGlastra I've had one situation (on this project) where the npm cache `~/.npm` was borked causing issues. Caches cannot be deleted in GitHub. The only way to fix it is to create a new cache and let the old one expire naturally. This takes care of that. Bump the number and a fresh set of caches are created.

.github/workflows/main.yaml
jleach: @TimoGlastra Done. Main will always build:

```yaml
on:
  pull_request:
    branches: [main]
    paths:
      - 'ios/**'
      - 'android/**'
      - '**/package-lock.json'
  push:
    branches: [main]
```

JamesKEbert (approved):
Looks great to me! Good work @jleach

PR #99, opened 2021-10-28
Fix VirtualizedLists nesting issue

### LINKED ISSUE #98: The VirtualizedLists component should not be embedded in a SafeAreaScrollView
As per the error message the app is throwing a `VirtualizedLists` should not be embedded in a `SafeAreaScrollView`.

```
VirtualizedLists should never be nested inside plain ScrollViews with the same orientation
because it can break windowing and other functionality - use another
VirtualizedList-backed container instead.
```

jleach:
# Summary of Changes

The `VirtualizedLists` does not like being nested in a `SafeAreaScrollView`. When done RN throws an error that is visually jarring. I fixed it by removing the `SafeAreaScrollView` and just using a plane old `View` because the two views this issue appears on don't need to scroll (AFAIK). On iOS usrs may be inclided to pull-to-refresh in this situation even though it has no effect.

# Related Issues

Fixes #98

# Pull Request Checklist


amanji (approved):

JamesKEbert:
Thanks for fixing this--I hadn't had time to hunt down this issue. Thanks!

PR #139, opened 2021-11-17
Post install script for installing Indy binary dependencies for Android

cvarjao:
# Summary of Changes

Externalize dependency for indy-sdk-react-native (android)

# Related Issues

N/A

# Pull Request Checklist

- [X] Updated **documentation** for changed code and new or modified features.


package.json
```diff
@@ -11,7 +11,8 @@
     "test": "jest",
     "lint": "eslint --ignore-path .gitignore --ext .js,.jsx,.ts,.tsx --color .",
     "prettier": "prettier --ignore-path .gitignore '**/*.+(js|jsx|json|ts|tsx|md|yml|yaml)'",
-    "postinstall": "patch-package",
+    "prepack": "rimraf *.tgz",
+    "postinstall": "patch-package && node post-install-setup.js",
```
TimoGlastra, line 15: This will run after every install right? Maybe you can add some form of check to see if this setup has already be done, that way on multiple runs it will actually only run once

TimoGlastra (commented):
I'm curious to hear a bit more about your setup. Are you consuming bifold as a library? How would that work? Do you import files/components from this package?

package.json
cvarjao: I think that can be optimized for sure. I am assuming `npm install` or `npm ci` is not something that developer runs very frequently, so I didn't think the overhead was too much. I've added to the existing postinstall, since I am not sure what `patch-package` is used for. I will add some extra checks to allow for caching

cvarjao:
> I'm curious to hear a bit more about your setup. Are you consuming bifold as a library? How would that work? Do you import files/components from this package?

The bootstrapping workflow that I am thinking does something like:
```
npx react-native init MyWallet --template @aries-framework/bifold-template
```
`@aries-framework/bifold-template` is package published in npm. That template will generate the whole project (it is a clone of [react-native-template-typescript](https://github.com/react-native-community/react-native-template-typescript) and in the `package.json` will have something like:
```
{
 "dependencies": {
    "@aries-framework/bifold": "^1.0.0"
  }
}
```

The generated/default index.js will looks something like:
```
 import Bifold from '@aries-framework/bifold'

 const Base = () => {
   useEffect(() => {
     LogBox.ignoreLogs(['VirtualizedLists', 'Mediator Invitation', 'Non-serializable values'])
   }, [])
 
   return (
     <NavigationContainer>
       <Bifold.App />
     </NavigationContainer>
   )
 }
```

cvarjao:
@TimoGlastra, [here](https://github.com/cvarjao/aries-bifold-template) is a crude idea of what I am thinking.

post-install-setup.js
```diff
@@ -0,0 +1,108 @@
...
+const {pathToFileURL} = require('url')
+const CWD = process.env.INIT_CWD || process.cwd()
+const JNI_LIBS_DIR = path.resolve(CWD, './android/app/src/main/jniLibs')
+
+/**
+ * Downloads file from remote HTTP[S] host and puts its contents to the
+ * specified location.
+ */
+ async function download(url, filePath) {
+    //fs.unlinkSync(filePath)
```
TimoGlastra:
```suggestion
```

post-install-setup.js
```diff
@@ -0,0 +1,108 @@
...
+}
+
+(async function() {
+    console.log(`Installing Android libraries at '${JNI_LIBS_DIR}'`)
+    for (const arch of archs) {
+        await downloadAndExtractIndyLib(arch)
+        await downloadAndExtractJnaLib(arch)
+    }
+    /* When bifold is used as a depency, copy ios framework embeded from within the package */
+    /* todo: figure out how to consume ios framework as a binary dependency as well (a la android)*/
```
TimoGlastra:
```suggestion
    /* When bifold is used as a dependency, copy ios framework embedded from within the package */
    /* todo: figure out how to consume ios framework as a binary dependency as well (a la android)*/
```

package.json
TimoGlastra: `patch-package` is used for patching files in the `node_modules` directory (see `patches` directory)

TimoGlastra (dismissed):
Haven't tested, but otherwise LGTM.

I'm still in favour of not adding this for the main repo as it complicates setup, but fine with merging.

@JamesKEbert what are your thoughts?

jleach:
I think this issue has gone stale. We can re-introduce it once we get the Indy SDK for Apple Silicon tested and in the bag.

PR #153, opened 2021-11-28
Added documentation aimed at security analysts, developers, and archi…

darrellodonnell:
# Summary of Changes

Created documentation files (markdown, PlantUML) primarily focused on security design considerations for apps based on Aries Bifold. 

# Related Issues

N/A

# Pull Request Checklist


docs/design/aries-rfcs.md
```diff
@@ -0,0 +1,33 @@
+# 0.0 Document Intent
+
+This document is intended to be consumed by non-developers so they can build an understanding of the underpinnings of the key cryptography usage in an Aries-based solution. In particular, the Aries Bifold wallet.
+
+# 1.0 Key ARIES RFCs
```
swcurran, line 5: I think this document should reference AIPs, not specific Aries RFCs.  It might call out specific Aries RFCs, but should do that indirectly, through the AIPs, since the AIPs are locked to a particular version of the RFC vs. to the "main branch" HEAD version.

This is highlighted by the reference to Aries RFC 0019, without noting which version (the commit linked to AIP 1.0 or 2.0?) and without referencing Aries RFC 0587, which is the V2 Encryption Envelope. For AIP 2.0, both should be supported, and there are features in 0587 for how both (and future) envelopes can be supported.

Since 0050 Wallet Design is not part of AIP 1.0 or 2.0, it is likely a good addition to this document, although I'm not yet familiar with that RFC.

darrellodonnell: I like the AIP referencing idea but I think we need to add the RFC references here as well. The goal, for analysts, is to gather all of the key references in one place and an AIP that references an RFC is indirect and makes for work for the analysts. We want to minimize the document traversal for them. Perhaps a top section that speaks to AIPs, the list the RFCs, with a note that the version of the RFCs needs to be locked to the AIP in question. thoughts?

swcurran: I'm OK with talking about the specific RFCs and their importance, but rather than linking directly to the RFC, link to the AIP and direct the reader to use the AIP link to find the referenced RFC.  Weird I know, and perhaps overkill, so I'm OK with this not being done. 

However, I do think it is CRUCIAL that AIPs be mentioned as a core part of this section.  We want people building to AIPs.

darrellodonnell: I think we need a quick discussion as the audiences that can get value from AIPs is different than at least one of the audiences that these documents are intended for (security analysts). I agree with the AIP addition, as AIPs are very helpful for developers and architects. The missing detail that answers "how does encryption happen on the wire" can't be an oblique reference. We're going to be adding pointers at the particular IETF RFCs in time, as an example, for where particular cryptographic approaches are normatively found.

darrellodonnell:
@jleach @TimoGlastra could one of you most awesome gents give this documentation only PR a quick review and approval?

TimoGlastra (approved):

TimoGlastra:
LGTM @darrellodonnell 

I left the PR as is as I'm no security expert

TimoGlastra:
Do you want me to merge?

darrellodonnell:
good to go

PR #170, opened 2022-01-11
fix: updated indy framework

### LINKED ISSUE #169: Bifold break after updating indy-sdk-react-native
Refer to that issue
https://github.com/hyperledger/indy-sdk-react-native/issues/29

Maybe we should temporary refer to tag "0.1.17"?

berendsliedrecht:
closes #169

adds the `proverDeleteCredential` function to the indy framework. I pulled this from my own wallet so I did not test it with the bifold project. However everything should work a-okay.


TimoGlastra:
@MosCD3 can you verify this fixes your issue?

ios/Pods/Frameworks/Indy.framework/Headers/IndyAnoncreds.h
```diff
@@ -415,6 +415,17 @@ And is documented in this HIPE:
...
+/**
+ Deletes credential by given id.
+ @param credId: Identifier by which requested credential is stored in the wallet
+ @param walletHandle Wallet handler (created by IndyWallet::openWalletWithName).
+ @param completion Callback that takes command result as parameter.
+ Returns void
+ */
++ (void)proverDeleteCredentialsWithId:(NSString *)credId
+                         walletHandle:(IndyHandle)walletHandle
+                           completion:(void (^)(NSError * error))completion;
```
MosCD3, line 427:
```suggestion
                           completion:(void (^)(NSError * error, NSString *anonString))completion;
```
You have to match the completion closure which is (Error?, String?) or the iOS will not build

MosCD3:
@TimoGlastra 
It will succeed and build after making the modification I proposed
That brings us to the ticket I created on Indy SDK repo
https://github.com/hyperledger/indy-sdk/issues/2459

I just need help from any one who was able to build the framework from the pod there, I was successful to build the core library but not the framework itself.

ios/Pods/Frameworks/Indy.framework/Headers/IndyAnoncreds.h
berendsliedrecht: Not too familiar with this, but it returns a void so this is correct I think.

MosCD3: It returns void but the closure takes in 2 parameters not 1

berendsliedrecht: Ah I see what is the issue. The indy-sdk-react-native uses `completionWithString` and it should be `completion`

will fix this right now.

TimoGlastra:
@blu3beri is this fixed?

berendsliedrecht:
> @blu3beri is this fixed?

I have tested it in another app and it works so I would assume so. @MosCD3 is it possible for you to test it?

MosCD3:
I tested on this app, on this repo and it doesn't work, it doesn't build 
@blu3beri  you said you would change the wrapper function in "indy-sdk-react-native" to use completion instead of "completionWithString", but I see the package is still at "0.1.14" 
For this to work Either change the package and update it here or silence the error by accepting my change which is passing an empty parameter to the function. 
I tested my suggestion and I was able to build and run bifold, so what's the harm? 
Even if you are planning to change "indy-sdk-react-native" to match the callback schema, you still could apply the quick fix and get new developers to build the app. Bifold on its public state is currently broken, test that yourself, download and try to build the app

TimoGlastra:
@MosCD3, berend has updated the package version. Could you tests with the latest version? We'd rather not manually modify the generated `Indy.Framework`

MosCD3:
@TimoGlastra  Now its working

TimoGlastra (approved):

PR #178, opened 2022-01-24
Fix PIN create and enter contrast

jleach:
# Summary of Changes

Because of the original two color pallet it was hard to change the theme and not run into contrast issues with the PIN entry screens. This change just better uses the updated pallet for a more uniform look and allows for more options when changing the color pallet.

I didn't change too much in the way of theme because these screens will be redone as Ontario comes up with a new and better way to use biometrics with a PIN override if not available.

IMG_5780

# Related Issues

n/a

# Pull Request Checklist

### DIFF
--- a/App/components/inputs/TextInput.tsx
+++ b/App/components/inputs/TextInput.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react'
-import { View, Text, StyleSheet, TextInput as TI, TextInputProps } from 'react-native'
+import { View, Text, StyleSheet, TextInput as RNTextInput, TextInputProps } from 'react-native'
 
-import { Colors, borderRadius } from '../../theme'
+import { Colors, TextTheme, borderRadius } from '../../theme'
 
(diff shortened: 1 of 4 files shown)


App/components/inputs/TextInput.tsx
```diff
@@ -1,26 +1,25 @@
...
-    margin: 2,
+    ...TextTheme.label,
+    marginBottom: 3,
   },
   textInput: {
     padding: 10,
     borderRadius,
     fontSize: 16,
-    backgroundColor: Colors.shadow,
+    backgroundColor: Colors.backgroundLight,
```
amanji: What would it look like if we left the background as `shadow` and changed the placeholder and input text colour?

jleach: Updated. I've added two greys that work well within the app. Screen shot above updated. Looks decent when changed a different color pallet also.

amanji (approved):

PR #179, opened 2022-01-25
Credentials list

amanji:
# Summary of Changes

* Applies theme variables through credentials list
* Fixes issue with displaying credential info
  - Updates schema parser helper to capitalize first letters of parsed schema names
  - Adds time issued to credential
* Fixes crash error when tapping to credential details screen

# Related Issues

N/A

# Pull Request Checklist


App/components/listItems/CredentialListItem.tsx
```diff
@@ -1,53 +1,52 @@
...
-import type { CredentialStackParams } from 'navigators/CredentialStack'
 
 import { useNavigation } from '@react-navigation/core'
 import { DateTime } from 'luxon'
 import React from 'react'
 import { View, StyleSheet, TouchableOpacity } from 'react-native'
 import Icon from 'react-native-vector-icons/MaterialIcons'
 
-import { Colors } from '../../theme'
+import { IndexedIndyCredentialMetadata, INDY_CREDENTIAL_KEY } from '../../constants'
```
jleach: Should we stick with one format or the the other? (using _ or not).

App/components/listItems/CredentialListItem.tsx
```diff
@@ -1,53 +1,52 @@
...
       onPress={() => navigation.navigate('Credential Details', { credentialId: credential.id })}
     >
       <View>
-        <Title style={{ color: Colors.text }}>{parseSchema(credential.metadata.schemaId)}</Title>
+        <Title style={{ color: Colors.text }}>
+          {parseSchema(credential.metadata.get<IndexedIndyCredentialMetadata>(INDY_CREDENTIAL_KEY)?.schemaId)}
+        </Title>
         <Text style={{ color: Colors.text }}>
-          Issued on {DateTime.fromJSDate(credential.createdAt).toFormat('LLL d, yyyy')}
+          Issued on {DateTime.fromJSDate(credential.createdAt).toFormat('LLL d, yyyy (hh:mm a)')}
```
jleach: Should we put the date format strings in constants so they can be used throughout the app?

App/screens/CredentialDetails.tsx
```diff
@@ -1,21 +1,49 @@
...
+    } catch (e: unknown) {
+      Toast.show({
+        type: 'error',
+        text1: (e as Error)?.message || t('Global.Failure'),
+      })
+      navigation.goBack()
+    }
+  }
+
+  const credential = getCredentialRecord(route?.params?.credentialId)
```
jleach: Over time we may need to move away from `?`: If there is a chance of the param not existing should we check and display an error?

App/theme.ts
```diff
@@ -112,7 +129,7 @@ export const Buttons = StyleSheet.create({
   primaryDisabled: {
     padding: 16,
     borderRadius: 4,
-    backgroundColor: 'rgba(66, 128, 62, 0.3)',
+    backgroundColor: Colors.darkGreenHeavyTransparent,
```
jleach, line 133: Should we avoid using color names in the theme? When we style to the BC Pallet will it make things more complicated (we don't have a green).

jleach (changes requested):
Have a read through the feedback Not all changes required, just food for thought.

App/components/listItems/CredentialListItem.tsx
amanji: Yeah I did think about this and for consistency I will align with using camel casing.

App/components/listItems/CredentialListItem.tsx
amanji: Good idea!

App/theme.ts
amanji: I went back and forth on this. My thoughts were to keep things DRY and avoid situations where we would have to make updates to multiple values. Instead of coming up colour names, is it better for us to stick to naming conventions more commonly used in style guides (ex primary, secondary, info, error, success, warning, etc)? It would also be nice to specify colour shades and opacities without having to use names like the one referenced here.

App/screens/CredentialDetails.tsx
amanji: We do check at the end whether the credential is defined but we can and should add an additional check for the existence of the route parameter further up.

App/theme.ts
amanji: Can we address this with theming?

jleach: Yup.

jleach (approved):

PR #203, opened 2022-02-09
Cleanup and update tests

jleach:
# Summary of Changes

- Fixed a snapshot and some light code cleanup.
- Forgot to commit (last PR) some accessibility cleanup to labels and make sure they exists on the on-boarding (Splash, Tutorial, Pin) screens.

# Related Issues

n/a

# Pull Request Checklist

### DIFF
--- a/App/components/shared/Pagination.tsx
+++ b/App/components/shared/Pagination.tsx
@@ -46,8 +46,8 @@ export const Pagination: React.FC<IPaginationProps> = ({
     }
   }
 
-  // The `TouchableOpacity` prop `disabled` is commented out below because it
-  // causes the `accessibilityLabel` to be permanently disappears when toggled.
+  // FIXME: Issue #204. Better to `disable` the `TouchableOpacity`
+  // controls rather than changing the color to transparent.
 
   return (
     <View style={style.pagerContainer}>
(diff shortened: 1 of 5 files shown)


App/components/shared/Pagination.tsx
```diff
@@ -46,13 +46,17 @@ export const Pagination: React.FC<IPaginationProps> = ({
     }
   }
 
+  // The `TouchableOpacity` prop `disabled` is commented out below because it
```
amanji: Will there ever be the need to disable the back button in lieu of the comment? If not then we can just remove the property along with the note.

jleach: There seems to be a bug where, if I disable a control then enable it the accessibility label does not show. My work around is to just change the colour to transparent - which I don't love. It causes the button to flicker on my phone. It's better to disable it. I was going to update RN in a later release and monitor to see if the problem goes away. The comment is more of a reminder for me to monitor the bug.

jleach: PS: I can remove it in favour of a GitHub issue.

amanji: Let's track it in a GH issue. For now just remove the `disabled` prop altogether and update the `FIXME` comment so devs know that it shouldn't be added.

jleach: Done. Issue #204 created to track.

amanji (approved):

PR #208, opened 2022-02-11
Work on connection notif

jleach:
# Summary of Changes

- Draft pattern for managing notifications via state changes;
- Add modal connection notifications.

# Related Issues

# Pull Request Checklist


App.tsx
```diff
@@ -63,6 +64,7 @@ const App = () => {
     <StoreProvider>
       <AgentProvider agent={agent}>
         <StatusBar barStyle="light-content" hidden={false} backgroundColor={Colors.primary} translucent={false} />
+        <ConnectionModal />
```
amanji, line 67: I'm generally liking the concept. Wondering if we could/would do the same for a Credential Offer or Proof Request? Perhaps not ATM since they are handled from the Home Screen. My only thought is whether we could be creating a tight coupling here with the Connection Modal. I can't think of any issues this could cause down the road right now.

amanji: To be honest this seems to create better separation than having that modal logic baked into the scan screen and would be pretty useful if we use different connection types. 👍

jleach: I think so. Most events are notably async so probably better to get some sort of background service dealing with that and the UI just focused on presentation when the state changes. AFJ helps quite a bit but I think there is still some room to manage the UI.

App/screens/Scan.tsx
```diff
@@ -86,7 +86,7 @@ const Scan: React.FC<ScanProps> = ({ navigation }) => {
 
     try {
       const url = event.data
-      if (isRedirecton(url)) {
+      if (isRedirection(url)) {
```
amanji, line 89: Good moving this to the helper utils and catching the spelling error 😅

amanji (approved):

PR #210, opened 2022-02-14
Adding wallet security services layer

MosCD3:
Signed-off-by: Mostafa <mostafa.youssef@ontario.ca>

# Summary of Changes

This is just a service layer that will serve the following purpose:

- Adding different security flows to initializing the wallet
- Using Argon key derivation function (KDF) to generate wallet key from user PIN
- Initializing different storage engines (Keychain/Prefs) to save generated keys/Ids

The service layer is supposed to be used within a global provider context or a reducer hook

Steps to do next in different PR

- Moving wallet init from App.tsx to be initialized upon successful pin entry 
- Wire wallet keys with wallet config before initializing the wallet

# Pull Request Checklist

- [] Updated **documentation** for changed code and new or modified features.


jleach:
@MosCD3 Able to sync `package.json` and `package-lock.json`? They need to be in sync for `npm ci` to work and our automated checks to run.

MosCD3:
@jleach Done

App/services/kdf.service.tsx
```diff
@@ -0,0 +1,8 @@
+import argon2 from 'react-native-argon2'
```
jleach, line 1: FYI: When uploading to the Apple or Google play store the app gets put on US servers. When a person downloads the app to install they are technically "exporting" the app from the USA. As a result, we're subject to cryptography export restrictions. This can be easy or complicated for a number of reasons, one being if you use standard crypto algorithms (NIST certified example) or bespoke roll-your-own solutions. We may collectively need to think through what it means to include Argon2 out of the box as it will likely have a notable impact on our regulatory process (ie. registering with the U.S. Department of Commerce’s Bureau of Industry and Security) 

From the Apple submission process:
Screen Shot 2022-01-06 at 10 58 06 AM

jleach (changes requested):
Would be good to bring up the use of `argon2` at the next community meeting. Also, it looks like a few got lined that should be ignored. These are generated by Xcode and I don't think should be committed back to the project:

- ios/Media.xcassets/AppIcon.appiconset/Contents.json
- ios/Media.xcassets/Contents.json

App/services/kdf.service.tsx
MosCD3: Argon2 is already implemented in indy anyways
https://github.com/hyperledger/indy-sdk/blob/master/wrappers/ios/libindy-pod/Indy/Wrapper/IndyWallet.h

MosCD3: ios/Media.xcassets/AppIcon.appiconset/Contents.json
ios/Media.xcassets/Contents.json

Those are the results of linting and formatting workflows done when committing the work, I don't think they should be ignored, they seems to have some formatting

jleach (dismissed):

amanji (approved):

PR #212, opened 2022-02-14
Add error modal

jleach:
# Summary of Changes

### Add test command

I added the command `npm run test:watch` to continuously run only the tests that change. Handy when working on tests or a component as its much faster. It also shows code coverage to let us know how throughout our test cases are.

### Add error modal

Add a modal to show error messages and a all to action as needed. Rather than use the modal throughout the app it uses the state management mechanics. Typical usage would be:

```JavaScript
const error = new BifoldError('Big Problem', 'What you did failed', 1024)
dispatch({
  type: DispatchAction.SetError,
  payload: [{ error }],
})
```

In this case, there is a title, message, and the error code would/could be what is used to better identify the problem for technical support. The above code would trigger displaying the message.

What the modal looks like:

IMG_A3ED672B868A-1

# Related Issues

n/a

# Pull Request Checklist


jleach:
@amanji This is the initial work on the error modal for situations where its a better fit than modals. I noticed that the notification list item and the error message notification are just a variation of style. Wondering if we should extract it to be its own component that gets embedded in a notification on the Home Screen as needed. We would have a few types: Success, Info, Warning and Error. Each would change the colours a bit. It would take in a few props:

- Title
- Message
- onCallToActionPressed

While in `draft` I called the above mentioned actions message box `BifoldErrorX.tsx`.
Thoughts?

jleach:
@amanji Reworked `BifildErrorX` to be `InfoTextBoxII`, a temporary name while a WIP. I think you'll get the idea that the text box itself is just a component that adapts its style a little to what type it is. It was either this or use `children` to fill in everything in the right column like title, body, error code (if present) and the button. Thoughts?

amanji (approved):

PR #220, opened 2022-02-23
Style the accept credentials screen

jleach:
# Summary of Changes

Updated the style on the accept credentials screen.

# Related Issues

n/a

# Pull Request Checklist


App/screens/CredentialOffer.tsx
```diff
@@ -31,123 +34,187 @@ const styles = StyleSheet.create({
...
+    paddingVertical: 15,
+    height: 90,
+  },
+  text: {
+    ...TextTheme.normal,
   },
 })
 
 const CredentialOffer: React.FC<CredentialOfferProps> = ({ navigation, route }) => {
+  if (!route || !route.params) {
```
amanji: Can usually just shorten to: `!route?.params`

App/screens/CredentialOffer.tsx
```diff
@@ -31,123 +34,187 @@ const styles = StyleSheet.create({
...
+    throw new Error('Unable to invitation from AFJ')
   }
 
   useEffect(() => {
+    if (credential.state === CredentialState.Declined) {
+      setDeclinedModalVisible(true)
+      return
+    }
+
     if (credential.state === CredentialState.CredentialReceived || credential.state === CredentialState.Done) {
```
amanji, line 97: Isn't it better to keep the logic separated in different effects?

__tests__/screens/CredentialOffer.test.tsx
```diff
@@ -1,126 +1,129 @@
...
-  beforeEach(() => {
-    jest.clearAllMocks()
+const credentialPath = path.join(__dirname, '../fixtures/degree-credential.json')
+const credential = JSON.parse(fs.readFileSync(credentialPath, 'utf8'))
 
-    testCredentialRecords[0].state = CredentialState.OfferReceived
+const connectionRecord = new ConnectionRecord(connection)
+const credentialRecord = new CredentialRecord(credential)
+// TODO:(jl) Make a fn to revive JSON dates properly and pass to `parse`
+credentialRecord.createdAt = new Date(credentialRecord.createdAt)
```
amanji, line 25: Can use the luxon package used in the app for parsing dates

App/screens/CredentialOffer.tsx
```diff
@@ -31,123 +34,187 @@ const styles = StyleSheet.create({
...
+
+  const { credentialId } = route.params
   const { agent } = useAgent()
   const { t } = useTranslation()
+  const [, dispatch] = useContext(Context)
   const [buttonsVisible, setButtonsVisible] = useState(true)
   const [pendingModalVisible, setPendingModalVisible] = useState(false)
   const [successModalVisible, setSuccessModalVisible] = useState(false)
   const [declinedModalVisible, setDeclinedModalVisible] = useState(false)
+  const dateFormatOptions = { year: 'numeric', month: 'short', day: 'numeric' }
```
amanji, line 71: Luxon is one of the packages currently used to parse dates. Not sure if we want to keep using that or refactor it out. There's a formatted string available in the `constants` module if using that package.

App/screens/CredentialOffer.tsx
```diff
@@ -31,123 +34,187 @@ const styles = StyleSheet.create({
...
-  if (!credential) {
-    Toast.show({
-      type: ToastType.Error,
-      text1: t('Global.Failure'),
-      text2: t('CredentialOffer.CredentialNotFound'),
-    })
-    navigation.goBack()
-    return null
+  if (!invitation) {
+    throw new Error('Unable to invitation from AFJ')
```
amanji, line 87: I would create a class called `AriesFrameworkError` or something like that to distinguish from `BifoldError` and some generic `Error`

__tests__/screens/CredentialOffer.test.tsx
jleach: I don't know if it will help here. Standard `Date` can parse ISO just fine. What's needed is a fn that can be passed to JSON.parse() to de-serialize date strings on the fly like `JSON.parse(data, fnToDetectAndDeserializeDates)`.

App/screens/CredentialOffer.tsx
jleach: My though is try and keep it simple where possible. The author of moment.js deprecated the package because most of what moment does can be done in native JS now days. If Luxton is just a different way to do things then its not adding much - maybe even detracting as new developers will need to learn Luxton.

amanji: Should we then remove it in favour of JS `Date`? I don't see value in maintaining two ways of doing something.

__tests__/screens/CredentialOffer.test.tsx
```diff
@@ -1,126 +1,129 @@
...
+    // @ts-ignore
+    useCredentialById.mockReturnValue(myCredential)
+
+    const tree = render(<CredentialOffer route={props as any} navigation={useNavigation()} />)
+
+    const onTheWayModal = tree.getByTestId('CredentialOffer.CredentialOnTheWay')
+    const declinedModal = tree.getByTestId('CredentialOffer.CredentialDeclined')
+    const addedModal = tree.getByTestId('CredentialOffer.CredentialAddedToYourWallet')
 
-    /**
```
amanji, line 91: I usually like to keep the acceptance criteria commented so we can see how the tests are structured to conform to those.

App/screens/CredentialOffer.tsx
jleach: I think so.

__tests__/screens/CredentialOffer.test.tsx
jleach: Ok. I'll see what I can add back in. My though on unit tests is they should run fast and be easy to maintain. I generally run `npm run test:watch` and am shooting for close to 80 code coverage including any key execution paths. Some times that won't necessarily align with acceptance criteria. Also, a snapshot test covers lots of ground and may cover one or more acceptance criteria (lots of bang for your buck).

App/screens/CredentialOffer.tsx
jleach: Good question. I though because they both operate on the same state `}, [credential])` they were related enough to go into the same hook. If they had different state dependencies, or cleanup requirements, I'd probably separate them. If you think they're better separated I'm 100% ok with that.

App/screens/CredentialOffer.tsx
jleach: Fixed

App/screens/CredentialOffer.tsx
jleach: I think they should become `BifoldError` with title, message, and code. My though is if any of these errors are thrown the offer screen won't be usable to the end user so it should throw an error the user can see and then exit/pop() back to the Home Screen. Or maybe these things should be checked before the offer screen is even shown. Thoughts? When I wrote that I was thinking these errors would be for developers only but maybe that's not the case. Could leave it as is and Bring the subject up at the next Bifold meeting. Would be good for feedback from AFJ developers.

amanji: The errors still are for developers but It would offer more granularity when logging ("was it a problem with the framework or the wallet itself?")

App/screens/CredentialOffer.tsx
amanji: The way I look at it, because they are independent conditional branches we lose nothing by separating them.

__tests__/screens/CredentialOffer.test.tsx
amanji: I agree on snapshots for testing presentation layout. I still think we should have functional tests that conform to the acceptance criteria that cover functional requirements. Ultimately that's what ensures conformity when the app is iterated on.

App/screens/CredentialOffer.tsx
jleach: I think unique "code" would help with that in the long run. In any case, I can create the new Error class.

__tests__/screens/CredentialOffer.test.tsx
jleach: Ok. I don't want to add them to my tests. Its just more work and makes them harder to maintain.

amanji (approved):

PR #263, opened 2022-03-31
Add Theme provider

cvarjao:
# Summary of Changes

- Add Theme provider
- Fixed lintstagedrc to run using npx so that there is no need to have dependencies installed globally (e.: json, tsx, typescript)
- Removed import aliasing (`components`)
- Added `typecheck` script so that we can use for typescript validation (`npm run typecheck`)

# Related Issues

# Pull Request Checklist

### DIFF
--- a/.lintstagedrc.json
+++ b/.lintstagedrc.json
@@ -1,3 +1,3 @@
 {
-  "**/*.{js,jsx,ts,tsx}": ["npm run lint -- --fix", "npm run prettier -- --write"]
+  "**/*.{js,jsx,ts,tsx}": ["npm run lint -- --fix", "npx prettier --write"]
 }
--- a/App.tsx
+++ b/App.tsx
@@ -16,14 +16,16 @@ import SplashScreen from 'react-native-splash-screen'
 import Toast from 'react-native-toast-message'
 
 import ConnectionModal from './App/components/modals/ConnectionModal'
+import ErrorModal from './App/components/modals/ErrorModal'
 import toastConfig from './App/components/toast/ToastConfig'
-import { initStoredLanguage } from './App/localization'
+import { initLanguages, initStoredLanguage, defaultTranslationResources } from './App/localization'
 import RootStack from './App/navigators/RootStack'
 import StoreProvider from './App/store/Store'
-import { ColorPallet } from './App/theme'
+import { defaultTheme as theme } from './App/theme'
+import { ThemeProvider } from './App/utils/themeContext'
 import indyLedgers from './configs/ledgers/indy'
 
-import ErrorModal from 'components/modals/ErrorModal'
+initLanguages(defaultTranslationResources)
 
 const App = () => {
   const [agent, setAgent] = useState<Agent | undefined>(undefined)
@@ -39,16 +41,18 @@ const App = () => {
   return (
     <StoreProvider>
       <AgentProvider agent={agent}>
-        <StatusBar
-          barStyle="light-content"
-          hidden={false}
-          backgroundColor={ColorPallet.brand.primary}
-          translucent={false}
-        />
-        <ConnectionModal />
-        <ErrorModal />
-        <RootStack setAgent={setAgent} />
-        <Toast topOffset={15} config={toastConfig} />
+        <ThemeProvider value={theme}>
+          <StatusBar
+            barStyle="light-content"
+            hidden={false}
+            backgroundColor={theme.ColorPallet.brand.primary}
+            translucent={false}
+          />
+          <ConnectionModal />
+          <ErrorModal />
+          <RootStack setAgent={setAgent} />
+          <Toast topOffset={15} config={toastConfig} />
+        </ThemeProvider>
       </AgentProvider>
     </StoreProvider>
   )
--- a/App/components/buttons/Button.tsx
+++ b/App/components/buttons/Button.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import { Text, TouchableOpacity } from 'react-native'
 
-import { Buttons, heavyOpacity } from '../../theme'
+import { useThemeContext } from '../../utils/themeContext'
 
 export enum ButtonType {
   Primary,
@@ -26,7 +26,7 @@ const Button: React.FC<ButtonProps> = ({
   disabled = false,
 }) => {
   const accessible = accessibilityLabel && accessibilityLabel !== '' ? true : false
-
+  const { Buttons, heavyOpacity } = useThemeContext()
   return (
     <TouchableOpacity
       onPress={onPress}
--- a/App/components/buttons/HeaderRight.tsx
+++ b/App/components/buttons/HeaderRight.tsx
@@ -2,7 +2,7 @@ import React from 'react'
 import { StyleSheet, Text, TouchableOpacity, View } from 'react-native'
 
 import Arrow from '../../assets/icons/large-arrow.svg'
-import { ColorPallet, TextTheme } from '../../theme'
+import { useThemeContext } from '../../utils/themeContext'
 
 interface HeaderButtonProps {
   title: string
@@ -12,27 +12,26 @@ interface HeaderButtonProps {
   disabled?: boolean
 }
 
-const style = StyleSheet.create({
-  container: {
-    flexDirection: 'row',
-    alignItems: 'center',
-  },
-  touchableArea: {
-    marginRight: 14,
-  },
-  title: {
-    ...TextTheme.label,
-    color: ColorPallet.grayscale.white,
-    marginRight: 4,
-  },
-  icon: {
-    transform: [{ rotate: '180deg' }],
-  },
-})
-
 const HeaderRight: React.FC<HeaderButtonProps> = ({ title, testID, accessibilityLabel, onPress, disabled = false }) => {
   const accessible = accessibilityLabel && accessibilityLabel !== '' ? true : false
-
+  const { ColorPallet, TextTheme } = useThemeContext()
+  const style = StyleSheet.create({
+    container: {
+      flexDirection: 'row',
+      alignItems: 'center',
+    },
+    touchableArea: {
+      marginRight: 14,
+    },
+    title: {
+      ...TextTheme.label,
+      color: ColorPallet.grayscale.white,
+      marginRight: 4,
+    },
+    icon: {
+      transform: [{ rotate: '180deg' }],
+    },
+  })
   return (
     <TouchableOpacity
       accessibilityLabel={accessibilityLabel}
--- a/App/components/inputs/CheckBoxRow.tsx
+++ b/App/components/inputs/CheckBoxRow.tsx
@@ -2,9 +2,9 @@ import React from 'react'
 import { View, StyleSheet, TouchableOpacity, Text } from 'react-native'
 import Icon from 'react-native-vector-icons/MaterialIcons'
 
-import { ColorPallet, TextTheme } from '../../theme'
+import { useThemeContext } from '../../utils/themeContext'
 
-interface Props {
+export interface Props {
   title: string
   accessibilityLabel?: string
   testID?: string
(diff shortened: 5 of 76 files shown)


cvarjao:
@JamesKEbert , @TelegramSam, would any of you have a chance to review this initial PR? I am holding off on the structural changes for the next PR

.lintstagedrc.json
```diff
@@ -1,3 +1,3 @@
 {
-  "**/*.{js,jsx,ts,tsx}": ["npm run lint -- --fix", "npm run prettier -- --write"]
+  "**/*.{js,jsx,ts,tsx}": ["npm run lint -- --fix", "npx prettier --write"]
```
jleach, line 2: Why this change only for prettier and not lint also?

App.tsx
```diff
@@ -16,14 +16,20 @@ import SplashScreen from 'react-native-splash-screen'
 import Toast from 'react-native-toast-message'
 
 import ConnectionModal from './App/components/modals/ConnectionModal'
+import ErrorModal from './App/components/modals/ErrorModal'
 import toastConfig from './App/components/toast/ToastConfig'
-import { initStoredLanguage } from './App/localization'
+import { initLanguages, initStoredLanguage, defaultTranslationResources } from './App/localization'
 import RootStack from './App/navigators/RootStack'
+import { pages as defaultOnboardingPages } from './App/screens/OnboardingPages'
+import { default as defaultSplashScreen } from './App/screens/Splash'
```
jleach: Do these need to be changed? Typically when we `export default Splash` you just need to `import Splash from ...`. Is it because you need a specific name?

App/components/buttons/Button.tsx
```diff
@@ -1,14 +1,14 @@
...
-import { Buttons, heavyOpacity } from '../../theme'
+import { useThemeContext } from '../../utils/themeContext'
 
 export enum ButtonType {
   Primary,
   Secondary,
 }
 
-interface ButtonProps {
+export interface ButtonProps {
```
jleach: Do these get imported elsewhere in the app now? I didn't think they needed to be exported.

App/components/inputs/SingleSelectBlock.tsx
```diff
@@ -2,25 +2,9 @@ import React, { useState } from 'react'
 import { StyleSheet, TouchableOpacity, View } from 'react-native'
 import Icon from 'react-native-vector-icons/MaterialIcons'
 
-import { borderRadius, ColorPallet } from '../../theme'
+import { Text } from '../'
```
jleach: How do you feel about using the full path rather than `../`? If I read this line and needed to change the Text component I'd hit Command-Shift-P in VSCode and open the file by name. With `../` I may have to lookup the file name or use the navigator if I didn't know it. I think it makes the code more readable.

App/screens/OnboardingPages.tsx
```diff
@@ -6,76 +6,88 @@ import CredentialList from '../assets/img/credential-list.svg'
...
-    marginRight: 20,
-    marginBottom: 10,
-  },
-  icon: {
-    marginRight: 10,
-  },
-})
-
-const customPages = (onTutorialCompleted: GenericFn) => {
+export function createStyle(theme: Theme) {
```
jleach: If you're refactoring in the future feel free to update these to the arrow notation `() => {}`. It's more uniform.

App/screens/ProofRequest.tsx
```diff
@@ -88,15 +86,23 @@ const ProofRequest: React.FC<ProofRequestProps> = ({ navigation, route }) => {
...
+        throw error
       }
-      setCredentials(creds)
-      setAttributeCredentials(Object.entries(creds?.requestedAttributes || {}))
     }
 
-    updateRetrievedCredentials(proof).catch(() => {
+    updateRetrievedCredentials(proof).catch((err) => {
+      // eslint-disable-next-line no-console
+      console.log(`ERROR2:`, JSON.stringify(err))
```
jleach: Able to remove this `console.log`?

App/utils/configurationContext.tsx
```diff
@@ -0,0 +1,25 @@
+import { createContext, useContext, FC } from 'react'
+
+import { pages } from '../screens/OnboardingPages'
+
+export interface ConfigurationContext {
+  onboarding: {
+    pages: typeof pages
+  }
+  splash: FC
```
jleach: I think it would be more readable to use `React.FC` here.

__tests__/components/AvatarView.test.tsx
```diff
@@ -2,9 +2,12 @@ import { render } from '@testing-library/react-native'
 import React from 'react'
 
 import AvatarView from '../../App/components/misc/AvatarView'
+import * as themeContext from '../../App/utils/themeContext' // note we're importing with a * to import all the exports
+import { defaultTheme } from '../../App/utils/themeContext'
 
 describe('AvatarView Component', () => {
   test('Renders correctly', async () => {
+    jest.spyOn(themeContext, 'useThemeContext').mockImplementation(() => defaultTheme)
```
jleach, line 10: Would it be easier to add a global mock in `__mock__`?

jleach (changes requested):
Most of my comments are not deal breakers for merging but I'd like to see the reply before approving and give others a chance to review so I marked my review as Request Changes.

While reviewing I was wondering if ti would be better to keep vars that don't change like `styles` out of the FC? I've always just put things that change inside and things that don't outside. Then you could just do something like `style.foo = { ...style.foo, ...{color: 'red}}` to override or use `style.compose()` to override inside the FC itself.

.lintstagedrc.json
cvarjao: That is the only one that was broken when I am running on my windows laptop. Didn't touch anything else that was already working

App.tsx
cvarjao: Either one works. No specific reason.  Happy to update

App/components/inputs/SingleSelectBlock.tsx
cvarjao: I am not to opinionated. The only problem is that aliases break when using as a library

App/screens/ProofRequest.tsx
cvarjao: Yes! left over.

App.tsx
JamesKEbert: +1 on this thought

App/components/inputs/SingleSelectBlock.tsx
JamesKEbert: So @jleach you're proposing `'../TextInput'` (which I am a fan of), but to confirm would this cause problems from a library perspective @cvarjao ?

App/components/listItems/NotificationListItem.tsx
```diff
@@ -7,12 +7,11 @@ import { useTranslation } from 'react-i18next'
...
 import Icon from 'react-native-vector-icons/MaterialIcons'
 
-import { TextTheme, ColorPallet } from '../../theme'
 import { GenericFn } from '../../types/fn'
 import { HomeStackParams, NotificationStackParams, Screens, Stacks } from '../../types/navigators'
 import { parsedSchema } from '../../utils/helpers'
-
-import Button, { ButtonType } from 'components/buttons/Button'
+import { useThemeContext } from '../../utils/themeContext'
+import Button, { ButtonType } from '..//buttons/Button'
```
JamesKEbert:
```suggestion
import Button, { ButtonType } from '../buttons/Button'
```
One too many slashes?

App/types/navigators.ts
```diff
@@ -72,6 +72,7 @@ export type CredentialStackParams = {
 export type HomeStackParams = {
   [Screens.Home]: undefined
   [Screens.Notifications]: undefined
+  [Screens.ProofRequestAttributeDetails]: { proofId: string; attributeName: string }
```
JamesKEbert, line 75: Maybe I'm missing something but this seems to be duplicated with line 90 below?

JamesKEbert (dismissed):
This looks good! I left pretty minor comments. Good work, thanks for working on this! :)

App/components/buttons/Button.tsx
cvarjao: Not for this PR, it will be needed eventually for the bifold as framework, but I guess we also need to have a conversation about how to implement models/interfaces that downstream projects can use. I will remove the export for now.

App/types/navigators.ts
cvarjao: Each stack have their own type, but they are very similar. There is probably an opportunity to refactor the code. But if it is alright, I will leave that for another work.

__tests__/components/AvatarView.test.tsx
cvarjao: I didn't want to mock the whole `themeContext.ts` component, only that particular function. I am not sure how to that using `__mock__`. I previously had in a [setup file](https://jestjs.io/docs/configuration#setupfiles-array) , but I didn't want to introduce that just yet on this PR

cvarjao: Also, `__tests__` are currently being ignored from typescript validation because there are a bunch of false positive errors. I think we will need to review and address  those those errors (possibly refactor) as another work.

App/components/inputs/SingleSelectBlock.tsx
amanji: I agree with being more explicit here for developer experience

amanji:
Wondering if we can further split this out into separate PRs, one for the config provider and one for the theme provider? Unless the changes need to be done together.

App/components/buttons/HeaderRight.tsx
```diff
@@ -12,27 +12,26 @@ interface HeaderButtonProps {
...
-  icon: {
-    transform: [{ rotate: '180deg' }],
-  },
-})
-
 const HeaderRight: React.FC<HeaderButtonProps> = ({ title, testID, accessibilityLabel, onPress, disabled = false }) => {
   const accessible = accessibilityLabel && accessibilityLabel !== '' ? true : false
-
+  const { ColorPallet, TextTheme } = useThemeContext()
+  const style = StyleSheet.create({
```
amanji, line 18: While I understand why we need to move the style configs into the component function I'm wondering if there is a better way to do this in order to keep function bodies lean and easier to read.

App/components/inputs/CheckBoxRow.tsx
```diff
@@ -2,31 +2,31 @@ import React from 'react'
 import { View, StyleSheet, TouchableOpacity, Text } from 'react-native'
 import Icon from 'react-native-vector-icons/MaterialIcons'
 
-import { ColorPallet, TextTheme } from '../../theme'
+import { useThemeContext } from '../../utils/themeContext'
 
-interface Props {
+export interface Props {
```
amanji: Export introduced

App/components/misc/QRScanner.tsx
```diff
@@ -5,48 +5,17 @@ import { useWindowDimensions, Vibration, View, StyleSheet, Text } from 'react-na
...
 import Icon from 'react-native-vector-icons/MaterialIcons'
 
-import { ColorPallet, TextTheme } from '../../theme'
-
-import QRScannerClose from 'components/misc/QRScannerClose'
-import QRScannerTorch from 'components/misc/QRScannerTorch'
-import { QrCodeScanError } from 'types/error'
+import { QrCodeScanError } from '../../types/error'
+import { useThemeContext } from '../../utils/themeContext'
+import QRScannerClose from '../misc/QRScannerClose'
```
amanji: These components are in the same directory so it may just be cleaner to use `./QRScannerClose`

App/components/misc/QRScannerTorch.tsx
```diff
@@ -2,34 +2,38 @@ import React from 'react'
...
-    borderRadius: 24,
-    marginBottom: 50,
-  },
-  icon: {
-    marginLeft: 2,
-    marginTop: 2,
-  },
-})
-
+function createStyles({ ColorPallet }: Theme) {
```
amanji, line 13: I like the idea used here to add styling without cluttering up the component function in reference to my other comment.

App/components/buttons/HeaderRight.tsx
amanji: Saw @jleach comment above and I agree with the sentiment of trying to keep things clean.

App/types/navigators.ts
amanji: This line should be removed. That screen should only be navigated to from a proof request details screen and hence is grouped as part of a different set of stack parameters. Having it as part of `HomeStackParams` is confusing.

App/components/inputs/CheckBoxRow.tsx
amanji: While we're at it we've been moving away from using the generic term `Props` and prefixing with the component name so this may be a good opportunity to rename to `CheckBoxRowProps`

App.tsx
jleach: @cvarjao Let's roll with the simplified notation:`import xxx from 'yyy'`

__tests__/components/AvatarView.test.tsx
jleach: Ok. We can leave it for now and I can show you some mocking tricks. I think you'll create a file in `__mock__` and use  `jest.requireActual()` or something then just mock what you want - not the entire thing.

jleach (dismissed):

amanji (dismissed):

jleach (approved):

amanji (approved):

PR #293, opened 2022-04-20
Added more specific variables in the theme file to allow for easier customization

ryanindicio:
# Summary of Changes

To help make Bifold more customizable, I tried to move a lot of the text styling and colors used throughout the app to their own individual assignments in the theme file. Please let me know if there is any feedback or ideas on how to improve these changes or make the naming / grouping more intuitive. 

Also included in this PR - updated ContactListItem styling to better reflect the new chat functionality, and added a line to auto run the post-install patches following an npm install. Also, made a small adjustment to the readme so new users know to add the .env file in the app directory. 

# Related Issues

n/a

# Pull Request Checklist

Pro Tip 🤓


app/index.js
```diff
@@ -2,50 +2,42 @@
 /**
  * @format
  */
-import 'react-native-gesture-handler'
```
amanji: Are we updating the linter to use double quotes and different spacing?

core/App/components/inputs/CheckBoxRow.tsx
```diff
@@ -38,9 +38,9 @@ const CheckBoxRow: React.FC<Props> = ({ title, accessibilityLabel, testID, check
         onPress={onPress}
       >
         {checked ? (
-          <Icon name={'check-box'} size={36} color={ColorPallet.brand.primary} />
+          <Icon name={'check-box'} size={36} color={Inputs.checkBoxColor.color} />
```
amanji, line 41: Minor nit: Do we want to mix case here? I'm ok with just using camelCase altogether

amanji (commented):
I think this is a good step to consolidate/clean up the styling. We in fact had a similar concept before moving towards using semantic names where possible. I think this makes sense now given how things can be overridden where necessary. The only thing I would maybe ask to change the quotation marks and spacing that was already in the repo.

core/App/components/inputs/CheckBoxRow.tsx
jleach: +1 For not mixing cases. Let's go with what the app primarily uses now.

app/index.js
jleach: +1 for keeping it consistent with single quotes.

ryanindicio: This totally makes sense, I'll update this. Not sure why my vscode did that

core/App/components/inputs/CheckBoxRow.tsx
ryanindicio: The existing style objects are using pascal case, did you want me to update those as well? (Buttons, ColorPallet, etc)

ryanindicio: So for instance with Buttons.primary, do we want the B to remain capitalized?

jleach: I think PascalCase for now. Can the changed to anything later so long as its consistent. I don't think any of the Theme mechanics use camelCase so let's avoid that for now as it introduces another inconsistency.

jleach:
@ryanindicio you'll need to update it against the base branch before we do a review otherwise GitHub will just ask for us to re-review.

jleach (approved):

PR #348, opened 2022-06-03
Amend license file to add Ontario team

MosCD3:
Signed-off-by: Mostafa <mostafa.youssef@ontario.ca>

# Summary of Changes

Amend license file to add Ontario team

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -2,181 +2,180 @@
                            Version 2.0, January 2004
                         http://www.apache.org/licenses/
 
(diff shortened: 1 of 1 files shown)


jleach (commented):
@MosCD3 What actually changed besides the addition of an author below?

jleach:
I don't think this is requred for any legal purposes and we should probably cleanup the Indecio entry. If this is a Linux Foundation / Hyperledger policy then sure - lets do it but I don't thin it is. For example, this pattern is not followed in ACA-py.

PR #369, opened 2022-07-10
Load genesis from GENESIS_URL

JonasNiestroj:
Signed-off-by: Jonas Niestroj <j.niestroj@protonmail.ch>

# Summary of Changes

The GENESIS_URL env variable seems to be not used at the moment. So i added the functionality for loading the genesis file and adding it to the indy ledgers.

I was not sure about making the ledger from the env a production ledger or not.

Maybe a json file with the format of 
`[{"name": "BCovrinTest", "isProduction": false, "genesisTransactions": ...}]`
would be more configurable then having default ledgers and adding one with the env file.

Feedback and other ideas appreciated.

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


TimoGlastra (commented):
I think I'd like the approach of setting the json better, as this will download the genesis file on every launch of the app.

JonasNiestroj:
@TimoGlastra I updated the PR with a new approach. Now its easy to update/add/remove new ledgers.
If this approach is suitable for you i will also update the README but before doing this, feedback is appreciated.

amanji:
I can no longer recall the exact convo about this, but we initially referenced the genesis txn strings via imports to keep the config compact and easy to read (some of them can get quite large). There was another convo about reintroducing the config back into the RootStack component rather than through an import, but I prefer the approach here for cleanliness and separation. I think @JamesKEbert had some comments about the config setup from before.

PR #407, opened 2022-08-10
feat: added connection notification to credential offer and proof presentations

wadeking98:

# Summary of Changes

Added a notification to the credential offer and the presentation request page (connection based). Notification lets user know that a new connection has been added to their contacts.

## Credential Offer Notification:
image
image

## Presentation Request Notification:
image
image

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


app/package.json
```diff
@@ -61,6 +61,7 @@
     "react-native-splash-screen": "3.3.0",
     "react-native-svg": "12.1.1",
     "react-native-toast-message": "2.0.1",
+    "react-native-unordered-list": "^1.0.4",
```
amanji: Wondering if we need to install a whole package for this? Doesn't seem like the one referenced is widely used either.

core/package.json
```diff
@@ -134,5 +134,8 @@
       "module",
       "typescript"
     ]
+  },
+  "dependencies": {
```
amanji: Should go as a peer dependency?

core/App/screens/ProofRequestAccept.tsx
```diff
@@ -73,6 +73,7 @@ const ProofRequestAccept: React.FC<ProofRequestAcceptProps> = ({ visible, proofI
   }
 
   useEffect(() => {
+    console.log(proof.state)
```
amanji: Remove console log statement.

amanji (changes requested):

core/App/screens/CredentialOffer.tsx
```diff
@@ -159,6 +160,7 @@ const CredentialOffer: React.FC<CredentialOfferProps> = ({ navigation, route })
             }}
           >
             {loading ? <RecordLoading /> : null}
+            <ConnectionNotification connectionID={credentialConnectionLabel} />
```
amanji: Do we want to check for the truthiness of the `credentialConnectionLabel` like what's done with the `proofConnectionLabel`?

core/App/screens/ProofRequest.tsx
```diff
@@ -253,6 +254,7 @@ const ProofRequest: React.FC<ProofRequestProps> = ({ navigation, route }) => {
             }}
           >
             {loading ? <RecordLoading /> : null}
+            {!!proofConnectionLabel && <ConnectionNotification connectionID={proofConnectionLabel} />}
```
amanji: Since an empty string is considered falsy we don't really need to explicitly convert it to boolean primitive with `!!`.

core/App/components/misc/InfoBox.tsx
```diff
@@ -177,9 +179,12 @@ const InfoBox: React.FC<BifoldErrorProps> = ({
         </Text>
       </View>
       <View style={styles.bodyContainer}>
-        <Text style={styles.bodyText} testID={testIdWithKey('BodyText')}>
-          {showDetails ? message : description}
-        </Text>
+        {bodyContent}
+        {(!!description || !!message) && (
```
amanji: If `description` or `message` is `undefined` it will be treated as a falsy value we don't really need to explicitly convert it to boolean primitive with `!!`.

core/App/components/misc/ConnectionNotification.tsx
```diff
@@ -0,0 +1,119 @@
...
+    notifyTextList: {
+      marginVertical: 6,
+    },
+    informationIcon: {
+      color: ColorPallet.brand.link,
+      marginLeft: 10,
+    },
+  })
+
+  const toggleInfoCard = async () => {
```
jleach: Does this need to to be `async`?

core/App/components/misc/ConnectionNotification.tsx
```diff
@@ -0,0 +1,119 @@
...
+    setInfoCardVisible(!infoCardVisible)
+  }
+
+  const goToSettings = async () => {
```
jleach: Does this need to to be `async`?

core/App/components/misc/InfoBox.tsx
jleach, line 183: Double not? I thin you just want `(description || message)`

core/App/components/misc/UnorderedList.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import { Text, View } from 'react-native'
+
+import { useTheme } from '../../contexts/theme'
+
+type ListProps = {
```
jleach: UnorderedListProps to follow convention.

jleach (changes requested):

core/App/screens/CredentialOffer.tsx
wadeking98: sometimes on the BC Wallet Demo the connection is formed but there's an error issuing the credential and the user has to restart the app and press the `resend credential` button on the demo. In this case the connectionID is sometimes blank even though a connection is formed. I could check for truthiness but I figured since a connection is always formed when getting a credential, the notification should always be on the credential offer page

core/App/screens/ProofRequest.tsx
wadeking98: strange, I was getting an error earlier when I didn't have the `!!`. I don't seem to be getting the error now though so maybe I was mistaken

core/App/components/misc/ConnectionNotification.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const [infoCardVisible, setInfoCardVisible] = useState(false)
+
+  const settingsNavigation = useNavigation<StackNavigationProp<RootStackParams>>()
+
+  const styles = StyleSheet.create({
+    modalCenter: {
+      flex: 1,
+      justifyContent: 'center',
+      alignItems: 'center',
+      backgroundColor: 'rgba(0,0,0,0.5)',
```
amanji: We haven't consistently applied an opaque background to the modal dialogs that I know of

wadeking98: I think the wire frames call for it here: https://xd.adobe.com/view/3db881d4-e963-42ec-a2c4-6b5ca23862a3-68eb/screen/a523d262-2549-4da9-9c2e-35ba1a84d7e8/ The background won't be opaque it will be translucent

amanji: Was mostly referring to the opacity setting for the background colour. I think maybe just confirm with the UI/UX team if we want to have this as a one off or should we use this as an opportunity to apply the same effect across all full screen modal dialogs like this.

wadeking98: @amanji Kim confirmed this is the desired design for the modal

knguyenBC: Please add a translucent backdrop consistently across all modals of this type. Thank you!

wadeking98: @knguyenBC by `this type` do you mean all connection notification modals? or all popup modals in general?

knguyenBC: @wadeking98 All popup modals in general

wadeking98: Got it. Could I do that update in a separate PR? I feel it might be a bit out of scope for this PR

knguyenBC: That sounds like a good plan.

knguyenBC (commented):
The only thing is the settings in the connection notification doesn't have a link to the settings like it does in the popup. However, the content for both the notification and the popup are subject to change so until then. It's good to go.

amanji (dismissed):

core/App/components/misc/ConnectionNotification.tsx
```diff
@@ -0,0 +1,119 @@
...
+    notifyTextList: {
+      marginVertical: 6,
+    },
+    informationIcon: {
+      color: ColorPallet.brand.link,
+      marginLeft: 10,
+    },
+  })
+
+  function toggleInfoCard() {
```
jleach: I think in JS and TS the fat-arrow fn should be used by default. It overcomes one of the more confusing aspects of JS and that is now `this` is scoped. The fat-arrow notation is more like what a developer would expect.

Since you're toggling a boolean state I think this will work:
```
  const toggleInfoCardVisibility = () => setInfoCardVisible((previousState) => !previousState)
```

core/App/components/misc/ConnectionNotification.tsx
```diff
@@ -0,0 +1,119 @@
...
+    setInfoCardVisible(!infoCardVisible)
+  }
+
+  function goToSettings() {
```
jleach: While I'd be more specific (navigateToSettings) with the name I have no problem with this.

jleach (changes requested):

jleach (approved):

PR #453, opened 2022-09-23
feat: initial work on logger service for Bifold

MosCD3:
Signed-off-by: Mostafa <moscd3@gmail.com>

In attempt to make the code more maintainable and encourage the public to collaborate, a logger service is important to trace major app flow and/or errors in the app, similar to AFJ
Thus is the service file only, following PR will wire up the service with a provider to use the logger anywhere in the App


jleach (changes requested):
Able to remove the word "service" from `log.services.ts`? Its already in a directory called "services" so I don't think including the word "service" in the file name adds additional clarity.

MosCD3:
> Able to remove the word "service" from `log.services.ts`? Its already in a directory called "services" so I don't think including the word "service" in the file name adds additional clarity.

renamed it to "logger.ts"

jleach (approved):

PR #455, opened 2022-09-27
feat: allow the user to toggle biometrics in settings

MosCD3:
Signed-off-by: Mostafa <moscd3@gmail.com>

# Summary of Changes

- Added flags and enums to support biometry toggle flow
- Added UI flags to use the same PinEnter screen for biometrics toggle
- Refactor navigation stack to include PinEnter in internal flows
- Added a new function in keychain service for disabling biometrics
- Refactored changePin function to support disabling biometrics because when disabling biometrics we want to return the wallet security key to re-save it with different flags

Re- Enable biometrics after user already chooses to disable biometrics at initial setup


jleach:
@MosCD3 Hey, Biometrics is enabled and working as expected. What 'ya adding with this one? I think @wadeking98 also implemented not storing the hash any longer and PIN fallback.

MosCD3:
Yes am just adding the option for the user to toggle between normal PIN flow and using biometrics after wallet is setup, currently the option is only available when you first create the Pin

amanji:
Sorry, merge conflicts introduced with #495

PR #502, opened 2022-10-21
feat: enable connection invites using shortened urls

thiagoromanos:
# Summary of Changes

This change enables connections using a url shortener request (see [https://github.com/hyperledger/aries-rfcs/tree/main/features/0434-outofband#url-shortening](https://github.com/hyperledger/aries-rfcs/tree/main/features/0434-outofband#url-shortening)).
To receive invites using a url shortener, I changed the call `agent?.oob.parseInvitation(uri)` to `agent?.oob.parseInvitationShortUrl(uri)`. They are basically the same, but the second one also has support to short urls. 

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/app/App.tsx
+++ b/app/App.tsx
@@ -15,7 +15,7 @@ import {
   NetInfo,
   defaultConfiguration,
 } from "aries-bifold";
-import React, { useEffect } from "react";
+import React, { useEffect, useMemo } from "react";
 import { StatusBar } from "react-native";
 import SplashScreen from "react-native-splash-screen";
 import Toast from "react-native-toast-message";
@@ -24,7 +24,10 @@ import Toast from "react-native-toast-message";
 initLanguages(translationResources);
 
 const App = () => {
-  initStoredLanguage();
+  
+  useMemo(() => {
+    initStoredLanguage().then();
+  }, [])
 
   useEffect(() => {
     // Hide the native splash / loading screen so that our
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -38,7 +38,7 @@
         "query-string": "7.0.1",
         "react": "17.0.2",
         "react-i18next": "11.13.0",
-        "react-native": "0.66.1",
+        "react-native": "0.66.5",
         "react-native-animated-pagination-dots": "0.1.72",
         "react-native-argon2": "2.0.1",
         "react-native-bouncy-checkbox": "^3.0.5",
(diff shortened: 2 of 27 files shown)


cvarjao:
Can you help me understand why we need to have an allow list of short URL domains? or am I misunderstanding it? I am wondering if there is another way to identify based on `c_i`/`d_m` parameters or some other way.

thiagoromanos:
> Can you help me understand why we need to have an allow list of short URL domains? or am I misunderstanding it? I am wondering if there is another way to identify based on `c_i`/`d_m` parameters or some other way.

The "url list" was a way I thought to know which ones are redirects and which are short urls. My idea was to, knowing the address I hosted my url shortener, I could tell the application to receive an invite instead of a redirect.
Looking into the code of aries-framework-javascript, [on the parseInvitation.ts file](https://github.com/hyperledger/aries-framework-javascript/blob/7be979a74b86c606db403c8df04cfc8be2aae249/packages/core/src/utils/parseInvitation.ts#L97), I noted that it checks if its out-of-band (`oob`) or an connection invitation (`c_i`, `d_m`), otherwise it considers an oob invitation from short url. Since the bifold checks if it is an redirect, I can't see other way to check short urls.
Is there any way to improve this solution? I'm out of ideas right now.

jleach:
@thiagoromanos Roughly paraphrasing @amanji here but he remarked that as per the OOB spec,  if a URL does not have the `c_i` or `d_m` query params then its a redirect. It looks like AFJ distinguishes between redirects and short URLs. @amanji thoughts on what to do? Opportunity to cleanup some code?


```
/**
 * Parses URL containing encoded invitation and returns invitation message. Compatible with
 * parsing short Urls
 *
 * @param invitationUrl URL containing encoded invitation
 *
 * @param dependencies Agent dependicies containing fetch
 *
 * @returns OutOfBandInvitation
 */
export const parseInvitationShortUrl = async (
  invitationUrl: string,
  dependencies: AgentDependencies
): Promise<OutOfBandInvitation> => {
  const parsedUrl = parseUrl(invitationUrl).query
  if (parsedUrl['oob']) {
    const outOfBandInvitation = OutOfBandInvitation.fromUrl(invitationUrl)
    return outOfBandInvitation
  } else if (parsedUrl['c_i'] || parsedUrl['d_m']) {
    const invitation = ConnectionInvitationMessage.fromUrl(invitationUrl)
    return convertToNewInvitation(invitation)
  } else {
    try {
      return oobInvitationFromShortUrl(await fetchShortUrl(invitationUrl, dependencies))
    } catch (error) {
      throw new AriesFrameworkError(
        'InvitationUrl is invalid. It needs to contain one, and only one, of the following parameters: `oob`, `c_i` or `d_m`, or be valid shortened URL'
      )
    }
  }
}
```

thiagoromanos:
@jleach and @cvarjao, I updated the PR with an update that might be better than my previous solution. Knowing that `parseInvitationShortUrl` will check for connection invites in the GET parameter of the url (`c_i`, `d_m` or `oob`) or shortened urls invites, and if it isn't one of those it throws an error, I used the `catch` block from `handleInvitation` to handle redirects. If this also fails, then a BifoldError is thrown.

jleach (approved):
Tested. Seems to be working fine. Nice work!

PR #508, opened 2022-10-26
feat: app commons provider with logger service implementation

MosCD3:
# Summary of Changes

This is a "Commons" provider meant to provide a shell for all globally common properties and methods needs to be accessed by the app modules that are not Auth/Aget ..etc something like a logger service and more to come on future refactoring

I also added and export function for logger service just to make it much easier to log directly without having to import the commons context... something like
`appLog(....)`

### DIFF
--- a/app/App.tsx
+++ b/app/App.tsx
@@ -1,5 +1,6 @@
 import {
   AgentProvider,
+  AppCommonsProvider,
   AuthProvider,
   ConfigurationProvider,
   NetworkProvider,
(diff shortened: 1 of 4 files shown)


jleach (dismissed):
LGTM. I'd consider being more specific with the name. It's fairly vague. Consider:

`CommonUtilProvider`
`CommonHelpers`
`UtilityProvider`
`HelperProvider`

 so it's clear what it provides.

amanji (dismissed):

amanji:
Can we remove references to `appLog`? It looks like the linters getting tripped up on that.

amanji:
@MosCD3 Are you able to fix the linter errors here so we can merge this?

amanji (approved):

PR #514, opened 2022-11-01
feat: allow state extension

jleach:
# Summary of Changes

This PR is the first step in allowing developers to extend the default state as needed.

# Related Issues

# Pull Request Checklist

Pro Tip 🤓


jleach:
@cvarjao @amanji I think this part is ready to merge. I've tested extending it and it seems to work fine. A little polishing perhaps on the Typescript types but I'm not sure what that looks like now. Only that when I merge the reducers I need to `// @ts-ignore` some ts complaints.

core/App/contexts/store.tsx
```diff
@@ -1,53 +1,96 @@
...
+import _defaultReducer, { ReducerAction } from './reducers/store'
+
+type Reducer = <S extends State>(state: S, action: ReducerAction) => S
+
+interface StoreProviderProps {
+  initialState?: State
+  reducer?: Reducer
+}
+
+export class DefaultBifoldState implements State {
```
amanji: Im not sure if I agree with converting to classes. Interfaces don't actually get compiled which means smaller builds and there's no overhead of instantiating.

amanji (changes requested):

core/App/index.ts
```diff
@@ -55,6 +56,8 @@ export type { GenericFn } from './types/fn'
 export type { AuthenticateStackParams } from './types/navigators'
 export type { OnboardingStyleSheet }
 export type { WalletSecret } from './types/security'
+// export type { State as DefaultState } from './types/state'
```
amanji: Dead code?

core/App/contexts/store.tsx
jleach: Let me see if I can revert it without TS complaining too much.

core/App/contexts/store.tsx
```diff
@@ -2,41 +2,49 @@ import React, { createContext, Dispatch, useContext, useReducer } from 'react'
...
+  },
+  credential: {
+    revoked: new Set(),
+    revokedMessageDismissed: new Set(),
+  },
+  error: null,
+  loading: false,
+}
+
+export const createInitialStateFactory = (): State => {
```
amanji: Why do we need this if we can just return `initialState`?

amanji (dismissed):

amanji (approved):

PR #527, opened 2022-11-11
feat: override scan screen

thiagoromanos:
Signed-off-by: Thiago Romano <thiagoromano.s@gmail.com>

# Summary of Changes

Allow the override of the scan screen

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/core/App/contexts/configuration.tsx
+++ b/core/App/contexts/configuration.tsx
@@ -3,6 +3,7 @@ import { createContext, useContext } from 'react'
 
 import { RecordProps } from '../components/record/Record'
 import OnboardingPages from '../screens/OnboardingPages'
+import { ScanProps } from '../screens/Scan'
 import { OCABundleResolver } from '../types/oca'
 import { SettingSection } from '../types/settings'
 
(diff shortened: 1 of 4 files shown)


cvarjao:
@thiagoromanos , I consider overwriting a whole screen a big hammer. This PR is fine, I was wondering if there anything more granular that could be made customizable. Again, I think this is a good PR as is. There might be great business reason why you want to overwrite.

thiagoromanos:
> @thiagoromanos , I consider overwriting a whole screen a big hammer. This PR is fine, I was wondering if there anything more granular that could be made customizable. Again, I think this is a good PR as is. There might be great business reason why you want to overwrite.

The reason is to allow the scanning of a custom qrcode (containing a JSON instead of an URL for example). Looking into the code, all the qrcode data handling is done in that screen, so I thought that there would be the place to override.
Another approach to this would be extract the qrcode handling from the screen to somewhere else and override that, but that would require some refactoring of the scan screen.
What you think would be the best option here?

jleach (approved):

PR #530, opened 2022-11-15
feat: nip validation complexity

jcdrouin21:
# Summary of Changes

This PR gives 6 separates level of complexity for the PIN. So, level 6 have every validations and level 1 contains only the validations from the Bifold. The PR also adds a helper for the creation of the PIN (the Helper is toggleable in the defaultConfiguration).

**Level 1:** Validates the length (6) and contains only numbers (Bifold by default)

**Level 2:** Validates that there are no repeats of 3 digits (e.g. 222) and that there is no sequence of three consecutive numbers (e.g. 234)

**Level 3:** Validates that there is no repeating two-digit sequence (e.g. 2525)

**Level 4:** Validates that there is no repetition of 2 digits (e.g. 22)

**Level 5:** Validates that there is no odd or even sequence (e.g. _02468)

**Level 6:** Validates that there is no cross pattern in the PIN (e.g. 753951)

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


cvarjao:
I am wondering if instead of "level" we just just have each rule being independently enabled/disabled. I can see a situation where someone might want to mix and match ... something like

Option 1:
```
interface PINRules {
   only_numbers: boolean,
   no_repeated_digits: boolean,  //or a number to say "by x consecutive digits" (by 1, or by 2 as proposed)
   no_consecutive_digits: boolean,
   no_odd_or_even_sequence: boolean,
   no_cross_pattern: boolean,
}
```

Option 2:
or they could also be independent validation rule/functions that are executed in the order provided
```
import * as rules from 'pin_rules'
rules =  [rules.only_numbers, rules.no_repeated_digits, rules.no_consecutive_digits, ...]
```
in this one you will need to define a function output type that returns the `outcome` (pass/fail) and maybe a reason of failure (e.g.: `message`)


Let's talk about it the next Aries User Group (Nov 22, 2022)

cvarjao (dismissed):

jleach (approved):

PR #543, opened 2022-12-01
feat: alert when no ledger service

jleach:
# Summary of Changes

In some corporate scenarios the wallet may be run behind a firewall that blocks ZMQ (ledger connections). This PR adds a check to connect to a ledger node to make sure it is accessable when a network conneciton is present.

# Related Issues

# Pull Request Checklist

Pro Tip 🤓


core/App/components/network/NetInfo.tsx
```diff
@@ -19,6 +19,20 @@ const NetInfo: React.FC = () => {
         text1: t('NetInfo.NoInternetConnectionTitle'),
       })
     }
+
+    if (isConnected) {
```
amanji, line 15: Can we just use an `else` statement in place if another conditional check?

core/package.json
```diff
@@ -135,5 +135,8 @@
       "module",
       "typescript"
     ]
+  },
+  "dependencies": {
+    "react-native-tcp-socket": "^6.0.3"
```
amanji: Should be added to `dependencies` in app and to `peerDependencies` in core?

core/App/contexts/network.tsx
```diff
@@ -34,16 +36,35 @@ export const NetworkProvider: React.FC = ({ children }) => {
...
 
+  const assertLedgerConnectivity = async (): Promise<boolean> => {
+    const nodes = fetchLedgerNodes()
+    const connections = []
+
+    if (typeof nodes === 'undefined' || nodes.length === 0) {
+      return false
+    }
+
+    for (const n of nodes) {
```
amanji: Can also do something like:

```
const connections = await Promise.all(nodes.map(n => canConnectToLedgerNode(n)));
return connections.includes(true);
```

jleach: Changed.

core/package.json
jleach: Fixed.

core/App/components/network/NetInfo.tsx
jleach: I fixed it to use an early return which I think cleans it up nicely.

core/configs/ledgers/indy/__mocks__/fixtures/sovrin-main-net-pool.json
```diff
@@ -0,0 +1,5 @@
+{
```
amanji, line 1: Do we want to be introducing test artifacts outside of the `__test__/` folder? Just concerned that it will get forgotten about.

jleach: Putting `__mocks__` where it is was suggested by the Jest docs. It seemed reasonable to put the data close to the code that uses it. Its not used by the tests but rather a faux ledger returned by the mock. I don't want to use real ones because they'll be updated periodically and break tests. I can move it inside `__mocks__`. Thoughts?

amanji (approved):

amanji (approved):

PR #562, opened 2022-12-12
fix: disable back button when updating biometry

wadeking98:

# Summary of Changes

Previously when updating the biometrics preferences through the settings screen, the user could sometimes click the back button before they were prompted to enter their pin. These changes disable the back button on the navigator after the user has updated their biometrics preferences, and re-enables the back button after the user has supplied their PIN.
## Before updating:
image

## After updating:
image

## After PIN:
image

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/core/App/contexts/reducers/store.ts
+++ b/core/App/contexts/reducers/store.ts
@@ -38,6 +38,7 @@ enum PrivacyDispatchAction {
 enum PreferencesDispatchAction {
   ENABLE_DEVELOPER_MODE = 'preferences/enableDeveloperMode',
   USE_BIOMETRY = 'preferences/useBiometry',
+  BIOMETRY_PREFERENCES_UPDATED = 'preferences/biometryPreferencesUpdated',
   PREFERENCES_UPDATED = 'preferences/preferencesStateLoaded',
 }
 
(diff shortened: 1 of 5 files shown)


jleach:
@wadeking98 Curious, why does it need another dispatch action?

wadeking98:
> @wadeking98 Curious, why does it need another dispatch action?

@jleach I needed to communicate with the settings navigation stack so I could disable/enable the back button. The best way to do that seemed to be through the state

jleach:
> > @wadeking98 Curious, why does it need another dispatch action?
> 
> @jleach I needed to communicate with the settings navigation stack so I could disable/enable the back button. The best way to do that seemed to be through the state

Wouldn't you be able to watch `useBiometry: false,`?

wadeking98:
> > > @wadeking98 Curious, why does it need another dispatch action?
> > 
> > 
> > @jleach I needed to communicate with the settings navigation stack so I could disable/enable the back button. The best way to do that seemed to be through the state
> 
> Wouldn't you be able to watch `useBiometry: false,`?

I can watch `useBiometry: false` but that value is in the store and is only updated once the user has successfully authenticated with their PIN. I need to know when the user presses the toggle button so that I can disable the back button until the user is presented with the PIN screen. The only state that gets updated exactly when the user presses the toggle button is `setCanSeeCheckPIN()` and I think that state is only at the `UseBiometry` component level so I can't watch it from the navigation stack

jleach (approved):

PR #576, opened 2022-12-20
feat: added exact time to issued date

wadeking98:

# Summary of Changes

Added exact time to issued attribute in credential details screen, this allows the user to know which credential is newest between multiple credentials issued on the same day.

Date translations are not working in react-native. Tested on IOS and Android.

image

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


jleach (changes requested):
The format used elsewhere in the app is `aDate.toLocaleDateString('en-CA', dateFormatOptions)}` where `dateFormatOptions` comes from constants. If you search for `toLocaleDateString` you'll fine examples. Fine to hard code `en-CA` for now but I think this needs to come from the current language of the phone. That way its in a ISO 8601 format and will show properly in French (et al):

```
> n.toLocaleDateString('fr-CA', { day: 'numeric', month: 'long',  year: 'numeric'})
'20 décembre 2022'
```

Also, consider extracting it to a fn that just returns the date with time. Not required though. Just an idea.

I think in Settings there is an example of how to get the current language from the phone. If not in the translation mechanics somewhere.

amanji:
I'd also consider getting the locale directly from `i18n` rather than passing hardcoded locales. I don't think we need to further abstract into another function since we're calling a method directly on the date object itself. But definitely make use of the `dateFormatOptions` in constants.

wadeking98:
right, the issue I'm running into with i18n is that it doesn't use the full locale string, just `fr` instead of `fr-CA`

wadeking98:
@amanji and @jleach, I'm running into issues with the locale date string in react native:
## javascript:
- input: `(new Date()).toLocaleDateString('fr', { day: 'numeric', month: 'long',  year: 'numeric'})`
- output: `21 décembre 2022`  
  
## react-native:
- input: `(new Date()).toLocaleDateString('fr', { day: 'numeric', month: 'long',  year: 'numeric'})`
- output: `December 21, 2022`  
  
I've tried with different locales too but it seems like react-native doesn't support locales on dates: https://stackoverflow.com/questions/51399551/why-isnt-tolocaledatestring-working-in-react-native-android

Judging by the comments it seems like it behaves differently on IOS than Android. We could consider switching to using another date library like moment

wadeking98:
Just tested with moment, still running into the same issue. It seems like the issue must be in the core of react-native

jleach:
@wadeking98 Interesting. Must be a lack of compatibility in the [Hermes](https://reactnative.dev/docs/hermes) JS engine. I suppose that is what it is. We can leave it then and do it the hard way. Although, I'm not sure how we get it to display properly when the phone is in French. I'll test it on iOS and see how it works. If it works on iOS as expected then maybe we can use the config change noted in the stack overflow article you cited. 

EDIT: Tested on iOS. Looks the same as Android. Used `new Date().toLocaleDateString('fr-CA', { day: 'numeric', month: 'long', year: 'numeric' })` got "January" expected "janvier". Its odd the post you mention has a comment that makes it seem like iOS works as expected.

amanji (commented):
I'm not sure if I agree with having this be a standalone implementation for date time formatting. While I get the need for localization I think we should look at being consistent with date representation.

amanji:
I would go so far to say that we create a helper function that outputs the right format and update every method call to `.toLocaleDateString` for consistency.

amanji (changes requested):
Please see my comments in the PR

wadeking98:
> I would go so far to say that we create a helper function that outputs the right format and update every method call to `.toLocaleDateString` for consistency.

@amanji should we just use something like moment then?

amanji:
I think so

PR #587, opened 2023-01-12
fix: add camera permission check to prevent black screen when scanner opened

bryce-mcmath:
# Summary of Changes

Added a unified camera permission check when scan is pressed. Previously, if a user selected the native camera permission modal option "Deny" or "Don't allow" they would be met with a black screen. Additionally, if they had already denied camera permission and pressed scan they would be again met with a black screen.

Here is the flow when a user presses scan for the first time:
permission_allow

Here is the flow when a user presses scan and has already given permission:
already_allowed

Here is the flow when a user presses scan and has already denied permission:
already_denied

Things to note:

Permission statuses: there are three main statuses for permissions; GRANTED, BLOCKED, and DENIED. DENIED means they have not granted permissions, but doesn't specify if they have actively refused the permission request or just haven't been asked yet. BLOCKED means they have been asked and have denied the request, but this status is only available through using the `request` method from `react-native-permissions`, rather than the `check` method. Using the `request` method will pop up the native permission modal if it hasn't already been BLOCKED. Hopefully that provides a bit of explanation for why the flow works the way it does.

Privacy state: the new CameraDisclosureModal replaces the old privacy policy popup screen (CameraDisclosure). Because of this, there is no need and no use of `didShowCameraDisclosure` state, so I've cut that out.

DismissiblePopupModal: the new flow required a dismissible popup modal, which I couldn't find an already-built component for, so I made this component which is only somewhat generalized but can easily be extended in the future if need be. It can be dismissed if the user taps outside of the modal or if they tap the clear icon. Its styles were adapted from a combination of the `InfoBox` and `PopupModal` components for consistency.

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


app/ios/Podfile
```diff
@@ -22,6 +22,10 @@ end
 target 'AriesBifold' do
   config = use_native_modules!
 
+  permissions_path = '../node_modules/react-native-permissions/ios'
```
jleach, line 25: Were these two lines added automatically or manually?

core/App/components/modals/CameraDisclosureModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+    },
+    buttonContainer: {
+      paddingTop: 10,
+    },
+  })
+
+  const onAllowTouched = async () => {
+    setRequestInProgress(true)
+    const granted = await requestCameraUse()
+    if (!granted) setShowSettingsPopup(true)
```
jleach: Can you use braces here? Its handy short hand but can lead to errors if someone less experienced picks up JS. One might be tempted to add a line below as part of the `if`.

core/App/components/modals/DismissiblePopupModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const { t } = useTranslation()
+  const { TextTheme, ColorPallet } = useTheme()
+  const iconSize = 30
+
+  const styles = StyleSheet.create({
+    modalCenter: {
+      flex: 1,
+      justifyContent: 'center',
+      alignItems: 'center',
+      backgroundColor: 'rgba(0,0,0,0.5)',
```
jleach: Curious if this is in the ColorPallet?

core/App/screens/Scan.tsx
```diff
@@ -54,17 +58,58 @@ const Scan: React.FC<ScanProps> = ({ navigation }) => {
     }
   }
 
-  const didDismissCameraDisclosure = () => {
-    dispatch({ type: DispatchAction.DID_SHOW_CAMERA_DISCLOSURE, payload: [] })
+  const permissionFlow = async (method: PermissionContract, permission: Permission) => {
```
jleach: You have a return type below in `requestCameraUse`, best to add one here for consistency.

jleach (changes requested):
Looks good. A few small comments.

app/ios/Podfile
bryce-mcmath: Manually, following the installation instructions here: https://github.com/zoontek/react-native-permissions#setup

core/App/components/modals/CameraDisclosureModal.tsx
bryce-mcmath: Will do

core/App/screens/Scan.tsx
bryce-mcmath: Will do

core/App/components/modals/DismissiblePopupModal.tsx
bryce-mcmath: Doesn't seem to be. I'll add it, since it's used in a few places now

jleach (approved):

PR #595, opened 2023-01-19
fix: added date translations

wadeking98:

# Summary of Changes

Added exact time to issued date.
Added time formats and month names to translation files since react-native doesn't support date locales.
Added moment library to support date formats

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


core/App/utils/helpers.ts
```diff
@@ -28,6 +30,51 @@ export function hashToRGBA(i: number) {
   return '#' + '00000'.substring(0, 6 - colour.length) + colour
 }
 
+export function formatTime(time: Date, translate: TFunction, params?: { long?: boolean; format?: string }): string {
```
jleach: Would creating a type for params help make sure you get the right combination?

```
export type FormatTimeParams = {long: boolean,  format?: undefined}  |
                    {long?: undefined,  format: string}
```

core/App/utils/helpers.ts
```diff
@@ -28,6 +30,51 @@ export function hashToRGBA(i: number) {
...
+    }
+  }
+  const customMonthFormat = formatString.match(customMonthFormatRe)?.[0]
+
+  let formattedTime = momentTime.format(formatString)
+
+  if (customMonthFormat) {
+    let monthReplacement = ''
+    const monthReplacementKey = momentTime.format(customMonthFormat)
+    if (customMonthFormat.length === 3) {
```
jleach: [Magic number](https://en.wikipedia.org/wiki/Magic_number_(programming))? Consider naming it with the rest of the consts at the top.

jleach (changes requested):

core/App/utils/helpers.ts
wadeking98: neither of the params are required. The format string isn't required because by default we grab it from the translations, and the `long` param isn't required because it will default to the short month 3 letter code

amanji: Not sure if we should be passing the translate function as a parameter. Can we not just import it here directly and use it?

amanji (changes requested):

core/App/components/listItems/ContactListItem.tsx
```diff
@@ -57,9 +57,7 @@ const ContactListItem: React.FC<Props> = ({ contact, navigation }) => {
       <View key={contact.id} style={styles.outerContainer}>
         <View style={styles.textContainer}>
           <Title style={ListItems.contactTitle}>{contact?.alias || contact?.theirLabel}</Title>
-          <Text style={ListItems.contactDate}>
-            {contact.createdAt.toLocaleDateString(i18n.language, dateFormatOptions)}
-          </Text>
+          <Text style={ListItems.contactDate}>{formatTime(contact.createdAt, t)}</Text>
```
amanji: The way I was thinking about this, is that the date and locale are passed in as parameters (and I guess the format options) everything else is encapsulated by the function.

core/App/utils/helpers.ts
wadeking98: No we can't because useTranslate() is a hook and hooks can't be used outside of components

core/App/components/listItems/ContactListItem.tsx
wadeking98: The locale is (kind of) being passed in by the translation function. We need to pass in the translation function because `useTranslations()` is a hook and we can't use hooks outside of components

core/App/utils/helpers.ts
amanji: We should be able to export the i18n instance to access the translation function directly from the localization module. Im not convinced that passing the translation function as a param is a good idea. It couples the formatter too closely to translation function.

core/App/components/listItems/ContactListItem.tsx
amanji: See my comment above. I'm of the opinion that the implementation/function signature should be as agnostic as possible to the npm packages

jleach (dismissed):

PR #611, opened 2023-02-02
fix: missing first credential

jleach:
# Summary of Changes

Addresses the issue of a delayed first credential by triggering mediator message pickup when a conneciton enters the "completed" state.

# Related Issues

# Pull Request Checklist

Pro Tip 🤓


core/App/screens/Connection.tsx
```diff
@@ -60,6 +72,14 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
     },
   })
 
+  useEffect(() => {
+    // FIX:(jl) There may be a better way to fetch queued messages.
+    // Under investigation.
+    if (connection && connection.state === 'completed') {
```
amanji: Wondering if we should also do this if the state is in `request-sent` as well? In both those scenarios, ACA-Py considers the connection established.

Otherwise this LGTM as a workaround.

jleach: Do you mean `response-sent`? If so I think completed is better because it is the eventual end state the connection gets to in AFJ. If we made the call during `response-sent` it would be called 2x almost back-to-back.

cvarjao (approved):
LGTM

PR #638, opened 2023-02-22
fix: adjust view to prevent inputs hidden by keyboard

bryce-mcmath:
# Summary of Changes

Allow dismissing of keyboard when tapping empty space outside of it and use `KeyboardAvoidingView` and `ScrollView` to make sure inputs can always be visible even with large text + zoom

Here is what it looks like with default settings on iOS:
create_pin_ios_normal

Here is what it looks like with max zoom and font on iOS:
create_pin_ios_zoom

Here is what it looks like with max zoom and font on Android:
create_pin_android_zoom

It looks good and functions properly on Android with default settings as well, I just didn't bother to make a gif of that. You'll have to trust me (or checkout this branch)

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/core/App/screens/PINCreate.tsx
+++ b/core/App/screens/PINCreate.tsx
@@ -13,6 +13,9 @@ import {
   TouchableOpacity,
   findNodeHandle,
   ScrollView,
+  KeyboardAvoidingView,
+  Platform,
+  TouchableWithoutFeedback,
 } from 'react-native'
 import { SafeAreaView } from 'react-native-safe-area-context'
 import Icon from 'react-native-vector-icons/MaterialIcons'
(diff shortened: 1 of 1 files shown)


wadeking98 (dismissed):

core/App/screens/PINCreate.tsx
```diff
@@ -119,100 +134,98 @@ const PINCreate: React.FC<PINCreateProps> = ({ setAuthenticated }) => {
...
-                if (PINTwoInputRef && PINTwoInputRef.current) {
-                  PINTwoInputRef.current.focus()
-                  // NOTE:(jl) `findNodeHandle` will be deprecated in React 18.
-                  // https://reactnative.dev/docs/new-architecture-library-intro#preparing-your-javascript-codebase-for-the-new-react-native-renderer-fabric
-                  const reactTag = findNodeHandle(PINTwoInputRef.current)
-                  if (reactTag) {
-                    AccessibilityInfo.setAccessibilityFocus(reactTag)
-                  }
+      <View style={[style.container]}>
+        <Text style={[TextTheme.normal, { marginBottom: 16 }]}>
```
jleach, line 140: Worth defining 16 and 24 above so they can be tweaked in once place rather than 2 or 4? If nobody would need to do this then I'm ok without the chage.

jleach (dismissed):

core/App/screens/PINCreate.tsx
bryce-mcmath: I've made the 24 into an `iconSize` variable above but I don't think I'll make the `marginBottom: 16` part of the stylesheet just because they aren't similar elements

jleach (approved):

PR #655, opened 2023-03-08
fix: updated support for display attributes

wadeking98:
# Summary of Changes

Added support for displaying placeholder attributes. Used for the person credential card screen:
image

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/core/App/components/misc/CredentialCard.tsx
+++ b/core/App/components/misc/CredentialCard.tsx
@@ -19,8 +19,7 @@ interface CredentialCardProps {
   onPress?: GenericFn
   style?: ViewStyle
   proof?: boolean
-  proofAttributes?: Attribute[]
-  proofPredicates?: Predicate[]
+  displayItems?: (Attribute | Predicate)[]
   existsInWallet?: boolean
 }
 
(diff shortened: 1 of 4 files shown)


core/App/components/misc/CredentialCard.tsx
```diff
@@ -43,19 +41,31 @@ const CredentialCard: React.FC<CredentialCardProps> = ({
     if (proof) {
```
jleach, line 41: Because these are all independent conditionals with a return do you think it would read better is separate if statements and use the "early return" pattern? I don't think the else-if syntax adds much and just makes it a bit harder to read.

```JavaScript
if (a) {
   return 1 + 1
}

if (b) {
   return 2 + 1
}

if (c) {
   return 3 + 1
}

return 5 + 5
}
```

General I find if / else better where the conditional is similar like

```
if (a.type === 'cat')
return 'meow'
else if (a.type === 'dog')
return 'howl'
else // for other types of a
return 'squeek'
```

jleach (changes requested):

jleach (approved):
Ty. I think that's a bit easier on the eyes.

PR #661, opened 2023-03-15
fix: grab credential name from proof restriction

### LINKED ISSUE #662: Incorrect credential name in proof if credential not found in wallet
As a holder, if the verifier issues me a proof request for a credential I don't have available in the wallet, I should be able to see the name of the credential being requested.  
  
currently the proof attribute group name is being used to display the credential name in bifold. This isn't guaranteed to be the same as the credential name.  
  
Parsing the credential name from the attribute/predicate proof restrictions is a more reliable method of getting the credential name.

wadeking98:
# Summary of Changes

Previously, if a credential for a proof request wasn't in the wallet, the wallet would display the requested attributes/predicate group name as the credential name. This functionality changes the behaviour so it first looks to see if there are any restrictions to cred_def_id, schema_name, or schema_id and grabs the name of the credential from there  

# Related Issues

Resolves: https://github.com/hyperledger/aries-mobile-agent-react-native/issues/662

# Pull Request Checklist

Pro Tip 🤓


bryce-mcmath (dismissed):

core/App/screens/ProofRequest.tsx
```diff
@@ -194,11 +196,21 @@ const ProofRequest: React.FC<ProofRequestProps> = ({ navigation, route }) => {
...
+        )
+        if (credNamesInAttrs.includes(startCase(credName))) {
+          credFound = true
+          return
+        }
+      })
+      return credFound
+    }
+
+    return typeof retrievedCredentials !== 'undefined' && Object.values(fields).every((c) => c.length > 0)
```
bryce-mcmath: I know you didn't write this line, but should we also be checking for `null`, not just undefined?

bryce-mcmath (commented):
Just one small question, otherwise still LGTM

core/App/screens/ProofRequest.tsx
wadeking98: we could use it as a falsy value and that way it would catch undefined, null, and the empty string

core/App/screens/ProofRequest.tsx
```diff
@@ -194,11 +196,21 @@ const ProofRequest: React.FC<ProofRequestProps> = ({ navigation, route }) => {
...
+        )
+        if (credNamesInAttrs.includes(startCase(credName))) {
+          credFound = true
+          return
+        }
+      })
+      return credFound
+    }
+
+    return !typeof retrievedCredentials && Object.values(fields).every((c) => c.length > 0)
```
bryce-mcmath: should this just be `return retrievedCredentials && Object.values(fields).every((c) => c.length > 0)`?

the `typeof` will give a non-empty string in almost every case, no?

bryce-mcmath (changes requested):

core/App/screens/ProofRequest.tsx
wadeking98: yes, my bad

wadeking98: okay fixed it

amanji (dismissed):

bryce-mcmath (approved):

PR #690, opened 2023-03-31
feat: improve modal screens themes and animated components

thiagoromanos:
# Summary of Changes

- Some adjustments were made on theme to better handle modal screens when those have backgroundColor different from the other screens.
- Static SVGs where moved to the theme, allowing to customize those images
- Animated components where moved to a new provider, so it can be customized as well. The reason I didn't use the theme for this customization is that some of those components use themed styles (it would have a cyclic dependency).

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


jleach:
@thiagoromanos  Able to fix the conflicts so we can review this one?

packages/legacy/core/App/components/misc/UnorderedListModal.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
```
wadeking98, line 1: What's the difference between `packages/legacy/core/App/components/misc/UnorderedListModal.tsx` and `packages/legacy/core/App/components/misc/UnorderedList.tsx`? they appear to be the same component to me

packages/legacy/core/App/components/modals/CommonRemoveModal.tsx
```diff
@@ -75,7 +75,7 @@ const CommonRemoveModal: React.FC<CommonRemoveModalProps> = ({ removeType, visib
     controlsContainer: {
       marginTop: 'auto',
       marginHorizontal: 20,
-      marginBottom: 108,
+      marginBottom: Platform.OS === 'ios' ? 108 : 20,
```
wadeking98, line 78: I'm curios what's going on here, does React-Native naturally render the margins differently on Android and IOS, and then this change compensates for that?

packages/legacy/core/App/components/views/FauxNavigationBar.tsx
```diff
@@ -9,7 +9,7 @@ import { GenericFn } from '../../types/fn'
 import { testIdWithKey } from '../../utils/testable'
 
 const defaultIconSize = 24
-const defaultStatusAndNavbarHeight = 44
+const defaultStatusAndNavbarHeight = Platform.OS === 'ios' ? 44 : undefined
```
wadeking98, line 12: Same thing here, just curious what the reason is for the difference in platforms

wadeking98:
looks good to me for the most part, once those few comments are addressed I'm ready to merge it

packages/legacy/core/App/components/misc/UnorderedListModal.tsx
thiagoromanos: It's the text color. If the moda bgcolor is different from the app's bgcolor, the text might be difficult to see in some cases. So, I put a different text color in the theme for modal screens. Since bifold has the same bgcolor for modals and the app, the text color will be the same. But in our app, it's different.

bryce-mcmath (commented):
Once Wade's comments are addressed I think this is good to merge

jleach:
> looks good to me for the most part, once those few comments are addressed I'm ready to merge it

@wadeking98 @thiagoromanos Wonder if a SaveArea would work?

packages/legacy/core/App/components/modals/CommonRemoveModal.tsx
thiagoromanos: Its because the big gap that android adds to the title (no idea why). Here is how it looks without that style:
unnamed (1)

packages/legacy/core/App/components/views/FauxNavigationBar.tsx
thiagoromanos: The same gap appears here. 
unnamed

thiagoromanos:
@wadeking98 , I marked as resolved just to track what I already answered. Feel free to comment it or unresolve those conversations

wadeking98:
> > looks good to me for the most part, once those few comments are addressed I'm ready to merge it
> 
> @wadeking98 @thiagoromanos Wonder if a SaveArea would work?

@jleach might work, but I think react native complains if you wrap a scrollview in a safeareaview, I think we'd have to change all our scrollviews to flatlists to get around the issue

wadeking98:
I think it looks good to me then, I'll approve but I won't merge yet in case the others have changes they suggest

wadeking98 (approved):

PR #793, opened 2023-05-30
fix: format verifier dates where possible

bryce-mcmath:
# Summary of Changes

Formatting dates in verifier where possible - I couldn't format the dates of parametrizable predicates, they had to remain numbers for proper functionality. Dates in proof requests themselves (as they appear to the holder) still need formatting, but that is for another PR.

Here is what the ProofRequestDetails screen looks like now:
Screenshot 2023-05-30 at 3 10 08 PM

Here is a parametrizable predicate (needs to remain a number):
Screenshot 2023-05-30 at 3 10 38 PM

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/screens/ProofRequestDetails.tsx
+++ b/packages/legacy/core/App/screens/ProofRequestDetails.tsx
@@ -1,6 +1,6 @@
 import { useAgent } from '@aries-framework/react-hooks'
 import { StackScreenProps } from '@react-navigation/stack'
-import React, { useCallback, useEffect, useState } from 'react'
+import React, { Dispatch, SetStateAction, useCallback, useEffect, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { FlatList, StyleSheet, Text, TextInput, View } from 'react-native'
 import { SafeAreaView } from 'react-native-safe-area-context'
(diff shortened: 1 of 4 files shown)


packages/legacy/core/App/screens/ProofRequestDetails.tsx
```diff
@@ -30,8 +31,31 @@ interface ProofRequestAttributesCardParams {
   onChangeValue: (schema: string, label: string, name: string, value: string) => void
 }
 
+const formatIfDate = (
```
jleach: Should this go into a helper lib with a unit test?

jleach (changes requested):
Didn't impact any tests which means it probably has none. Not a blocker but if you can add one or two it would be a good pattern to get into. Consider moving the helper fn and writing a test for it though. Its a good quick win.

packages/legacy/core/App/screens/ProofRequestDetails.tsx
bryce-mcmath: done

jleach (approved):

amanji:
Is it possible to use something like a date selector if we know the field will be a `dateint`?

PR #813, opened 2023-06-09
fix: migrating wallet to askar when used for auth

jleach:
# Summary of Changes

- Merge in changes from main;
- Add support to migrate the wallet to ASKAR when its done for authentication (before initialization takes place). The changes for this change are in `migrate.ts` and `auth.tsx` with the function from migrate added into `Splash.tsx`. 

# Related Issues

n/a

# Pull Request Checklist

Pro Tip 🤓


jleach:
@TimoGlastra I merged in main and fixed conflicts. The main changes to fix migration are in `migrate.ts` and `auth.tsx` with the function from migrate added into `Splash.tsx` also. LMK what you think. If you merge this into [hyperledger:feat/upgrading-to-AFJ-040](https://github.com/hyperledger/aries-mobile-agent-react-native/tree/feat/upgrading-to-AFJ-040) then move it out of "Draft" and we'll review and test tomorrow.

packages/legacy/core/App/contexts/auth.tsx
```diff
@@ -85,14 +86,22 @@ export const AuthProvider: React.FC = ({ children }) => {
...
+        return false
+      }
+
+      const hash = await hashPIN(PIN, secret.salt)
+
+      if (!store.migration.didMigrateToAskar) {
+        await migrateToAskar(secret.id, hash)
+        dispatch({
+          type: DispatchAction.DID_MIGRATE_TO_ASKAR,
+        })
```
TimoGlastra, line 102: We're calling this dispatch twice now. Doesn't that make it skip the actual wallet migration?

TimoGlastra (commented):
One comment on the DID_MIGRATE_TO_ASKAR dispatch being called twice for different steps

packages/legacy/core/App/contexts/auth.tsx
jleach: I think its needed because we do two different authentications methods. If the PIN is used, we need to upgrade the wallet then check the PIN actually unlocks it. If biometrics is used, then we know we have the key to unlock the wallet, so we just upgrade and initialize the agent.

TimoGlastra: Yeah but if we call disptach it will update the state right, and then when it chrcks in the splash.tsx file the state is already updated?

jleach: @TimoGlastra In splash.tsx it's gated on L218 so the logic won't run if the wallet is already updated:

```
if (!didMigrateToAskar(store.migration)) {
   ...
}
```

TimoGlastra: didMigrateToAskar is triggered based on the state. If we already call the dispatch when we migrate the pin check wallet, then this will see it as if we're already migrated.

The gate checks the state only

jleach: @TimoGlastra Is there a way to check with the wallet if its upgraded already or not? Maybe I can add something to the migration helper to check if its updated already then move on - nothing to do. I'm looking for the recommended approach. Ty.

jleach: @TimoGlastra Ok, here are two options. I think we could do both if you feel strongly about (2). We move ahead with this PR and get #666 merged as is because the db cannot be upgraded twice. Then move ahead with (2) as it will take some time to snake through all the different repos.

1. Do Nothing

The first line of the migration code [L133](https://github.com/hyperledger/aries-askar/blob/main/src/migration/mod.rs) in Rust has a check to make sure an updated DB is not updated a second time. We could just wrap the checks in JS and catch this error. Although, we'd have to test the string error message to see what type of error  it is because there is no code:

```
        if self.is_migrated().await? {
            self.close().await?;
            return Err(err_msg!(Backend, "Database is already migrated"));
        }
 ```

2. Expose `is_migrated()`

I can put a PR in to the rust libs to make `is_migrated()` public and expose it via the wrapper. We could then remove the state from PR 666 and use this method to check if the db is upgraded.

TimoGlastra: Couldn't we add a separate paramter to the state that keeps whether the pin is migrated?

PR #815, opened 2023-06-09
fix: use common remove for with-credentials

bryce-mcmath:
# Summary of Changes

Previously if a user tried to delete a contact that had credentials in their wallet, it would show the regular remove modal and then only if they attempted to fully delete it, would they see the modal for not being able to delete a contact (and only after they dismissed the first modal.) I refactored the separate modals into just the one `CommonRemoveModal` and adjusted the logic so that they display appropriately. Also added test for the updated `CommonRemoveModal` option.

Here's what the contact remove flow looks like now:
remove_contact_with_credentials

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/components/modals/CommonRemoveModal.tsx
+++ b/packages/legacy/core/App/components/modals/CommonRemoveModal.tsx
@@ -133,6 +133,10 @@ const CommonRemoveModal: React.FC<CommonRemoveModalProps> = ({ usage, visible, o
       return t('ContactDetails.RemoveContact')
     }
 
+    if (usage === ModalUsage.ContactRemoveWithCredentials) {
+      return t('ContactDetails.GoToCredentials')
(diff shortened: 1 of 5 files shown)


packages/legacy/core/App/components/modals/CommonRemoveModal.tsx
```diff
@@ -133,6 +133,10 @@ const CommonRemoveModal: React.FC<CommonRemoveModalProps> = ({ usage, visible, o
       return t('ContactDetails.RemoveContact')
     }
 
+    if (usage === ModalUsage.ContactRemoveWithCredentials) {
```
jleach, line 136: Wondering if we should change to a `switch` statement since its all the same var?

jleach (approved):

PR #818, opened 2023-06-12
fix: handled startCase() issue on some cred names

wadeking98:
# Summary of Changes

Previously the wallet would show error branding on the credential `Demo (For Demo)` this is because the wallet performs a check to see if the requested credential exists in the wallet. It casts the credential name to `Start Case`: eg. `hello_world` => `Hello World`. In IOS the start case of `Demo (For Demo)` is `Demo For Demo` with the brackets removed so the wallet branding thinks it doesn't exist in the wallet. I've changed the logic to check for start case or plain text, this resolves the issue. 

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/screens/ProofRequest.tsx
+++ b/packages/legacy/core/App/screens/ProofRequest.tsx
@@ -12,7 +12,6 @@ import {
   GetFormatDataReturn,
 } from '@aries-framework/core/build/modules/proofs/models/ProofServiceOptions'
 import { useAgent, useConnectionById, useProofById, useCredentials } from '@aries-framework/react-hooks'
-import startCase from 'lodash.startcase'
 import React, { useState, useMemo, useEffect } from 'react'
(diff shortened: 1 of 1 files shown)


jleach (dismissed):

cvarjao:
@wadeking98 , if AFJ is already matching and returning credentials that matches the proof request, why do we need to check for credential name? I think we need to be careful with credential naming matchings like that. It might be a vector for allowing match of different names: `Demo (For Demo`, `Demo For Demo)`, `Demo (For Demo)`, `Demo For Demo`, seems to all resolve to the same thing. It is okay to present that in the UX, but it is not a good data for testing equality. Analogous to javascript world, Sometimes "loose equality" (`==`) is okay, sometimes we need to use "strict equality" (`===`)

jleach (approved):

jleach (approved):

PR #850, opened 2023-06-20
fix: modal status bar

wadeking98:
# Summary of Changes

Added status bar background to IOS. IOS doesn't really have a notion of status bar background colour, the IOS status bar is always transparent. I added a simple view at the top of the modal to give the effect of a coloured status bar. I added the component to the usebiometry in the settings as well as the camera disclosure modal
Screen Shot 2023-06-20 at 10 39 33 AM

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/components/misc/IOSStatusBar.tsx
+++ b/packages/legacy/core/App/components/misc/IOSStatusBar.tsx
@@ -1,0 +1,17 @@
+import React from 'react'
+import { Platform, View } from 'react-native'
+
+import { useTheme } from '../../contexts/theme'
+
+
+
+const IOSStatusBar: React.FC = ({ ...props }) => {
+    const { ColorPallet } = useTheme()
+    return (
+        <>
+            {Platform.OS === "ios" && <View style={{ height: 40, backgroundColor: ColorPallet.brand.primary }} {...props}></View>}
+        </>
+    )
+}
+
+export default IOSStatusBar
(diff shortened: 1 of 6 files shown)


packages/legacy/core/App/components/misc/IOSStatusBar.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import { useTheme } from '../../contexts/theme'
+
+
+
+const IOSStatusBar: React.FC = ({ ...props }) => {
+    const { ColorPallet } = useTheme()
+    return (
+        <>
+            {Platform.OS === "ios" && <View style={{ height: 40, backgroundColor: ColorPallet.brand.primary }} {...props}></View>}
```
bryce-mcmath: Is there a calculation we can use instead of just 40 pixels? Thinking about different phones

bryce-mcmath (changes requested):

jleach:
@wadeking98 I obvs don't have this PR merged in but I see a status bar already. Is this more specific to a phone size?

Screenshot 2023-06-20 at 1 52 24 PM

jleach:
@wadeking98 @bryce-mcmath Should there be a status bar? Its a modal which usually does not get the top status bar or nav bar - no? We could just change it to be dark themed so the text is black.

wadeking98:
> @wadeking98 @bryce-mcmath Should there be a status bar? Its a modal which usually does not get the top status bar or nav bar - no? We could just change it to be dark themed so the text is black.

I prefer that approach, it seems cleaner

PR #868, opened 2023-06-30
feat: new scan workflow

jleach:
# Summary of Changes

We got some feedback letting is know the existing workflow of automatically waiting for, and processing offers and proofs was not awesome. Based on the feedback, we don't automatically process these unless its a proof request packaged with an OOB connection. Otherwise you'll be navigated to the contact details of the connection that was established.

We'll re-visit this when goal codes are implemented and use them to determine what logic to execute. 

I wasn't sure how much dead code to cleanup since some or all of it will be used again in the figure. My initial though was to remove it all - and I'm happy do do so based on feedback.

# Related Issues

bcgov/bc-wallet-mobile#1141

# Pull Request Checklist

Pro Tip 🤓


jleach:
@bryce-mcmath @wadeking98 Thoughts on what to do with the dead code? My though is it will be used again in the future but I'd prefer to delete it because I have no idea when it will be implemented as part of the goal codes ticket.

packages/legacy/core/App/screens/Connection.tsx
```diff
@@ -72,102 +69,98 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
...
-      }
+      // No goal code, we don't know what to expect next,
+      // navigate to the contact details.
+      navigation
+        .getParent()
+        ?.navigate(TabStacks.HomeStack, { screen: Screens.ContactDetails, params: { connectionId } })
+    }
+
+    return () => {
+      connection === undefined
```
bryce-mcmath: Just curious, what is the purpose of returning a boolean from the cleanup function?

packages/legacy/core/App/screens/Connection.tsx
```diff
@@ -72,102 +69,98 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
...
+          navigation.navigate(Screens.ProofRequest, { proofId: notification.id })
+          dispatch({ isVisible: false })
           break
         }
       }
     }
-  }, [notifications, state])
+
+    return () => {
+      notifications === undefined
```
bryce-mcmath: Same as above

bryce-mcmath (commented):
Just a couple comments out of curiosity

packages/legacy/core/App/screens/Connection.tsx
jleach: I'm testing to see if it clears up a memory leak waring from metro. My though was the connection or notification below change after the screen is dismissed triggering a useEffect and causing the warning. Might be a long shot but I'm not seeing it ATM.

packages/legacy/core/App/screens/Connection.tsx
jleach: See above.

packages/legacy/core/App/screens/Connection.tsx
bryce-mcmath: But isn't that a equality operator and not an assignment? I think those two vars are declared with `const` too. But if it's still working somehow we can keep it, that works for me

jleach: Oh, yes, good catch.

bryce-mcmath (commented):
Are we still changing it to navigate to the chat screen rather than the contact details screen?

packages/legacy/core/App/screens/Connection.tsx
jleach: Removed. Wasn't what I was after.

packages/legacy/core/App/screens/Connection.tsx
jleach: Removed. Wasn't what I was after.

jleach:
> Are we still changing it to navigate to the chat screen rather than the contact details screen?

Yes, this navigates to the chat screen now.

packages/legacy/core/App/screens/Connection.tsx
```diff
@@ -72,102 +69,98 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
...
-          navigation.navigate(Screens.CredentialOffer, { credentialId: notificationRecord.id })
-          break
-        case 'ProofRecord':
-          navigation.navigate(Screens.ProofRequest, { proofId: notificationRecord.id })
-          break
-        default:
-          throw new Error('Unhandled notification type')
-      }
+      // No goal code, we don't know what to expect next,
+      // navigate to the contact details.
```
bryce-mcmath: It really looks like it still navigates to the contact details screen

jleach: Yes, didn't push my change + I just updated the docs. Should be fixed now.

bryce-mcmath (approved):

PR #872, opened 2023-07-04
fix: make sure walletName is not undefined when fetching from ls

bryce-mcmath:
# Summary of Changes

Previously, if an older instance of the wallet that had already gone through onboarding was initialized it would have an undefined walletName in preferences, which would then overwrite the default value. This caused errors in connections because the label was missing. This fix is two-pronged, first it will convert undefined `walletName` values in localstorage to `generateRandomWalletName` when preferences are loaded, then it will use a backup value of `Aries Bifold` if `walletName` is still somehow undefined.

EDIT: I also removed the check for the wallet being named from the onboarding completeness check, so that older already-initialized wallets don't end up having to go back to onboarding. (Something I did for BC Wallet already but not for Bifold)

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/screens/Splash.tsx
+++ b/packages/legacy/core/App/screens/Splash.tsx
@@ -30,14 +30,8 @@ import {
 import { getAgentModules, createLinkSecretIfRequired } from '../utils/agent'
 import { migrateToAskar, didMigrateToAskar } from '../utils/migration'
 
-const onboardingComplete = (state: StoreOnboardingState, enableWalletNaming: boolean | undefined): boolean => {
-  return (
-    state.didCompleteTutorial &&
-    state.didAgreeToTerms &&
-    state.didCreatePIN &&
-    (state.didNameWallet || !enableWalletNaming) &&
-    state.didConsiderBiometry
-  )
+const onboardingComplete = (state: StoreOnboardingState): boolean => {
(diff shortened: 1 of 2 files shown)


packages/legacy/core/App/screens/Splash.tsx
```diff
@@ -214,7 +208,7 @@ const Splash: React.FC = () => {
 
         const newAgent = new Agent({
           config: {
-            label: store.preferences.walletName,
+            label: store.preferences.walletName || 'Aries Bifold',
```
cvarjao, line 211: would it be better to set the wallet name if the preference is undefined before initializing/loading the wallet?
```
if (!store.preferences.walletName) {
      dispatch({
        type: DispatchAction.UPDATE_WALLET_NAME,
        payload: [generateRandomWalletName()],
      })
  // retrieve updated `store`?
}

const newAgent = new Agent({ ... })
```

bryce-mcmath: Line 211 is just an additional backup in case somehow it's undefined after setting preferences from localstorage, but if it is undefined when preferences are being fetched it is being saved as `generateRandomWalletName()` in that action. So in the flow of an older wallet running the new code it will be 'My Wallet - ####' not 'Aries Bifold'.

wadeking98 (approved):

PR #874, opened 2023-07-06
fix: verifier notifications

wadeking98:
# Summary of Changes

Previously I removed mobile verifier notifications in the home screen. This change adds them back in as well as fixes bugs while viewing or deleting the notification.
image
image

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/components/listItems/NotificationListItem.tsx
+++ b/packages/legacy/core/App/components/listItems/NotificationListItem.tsx
@@ -1,6 +1,5 @@
-import type { CredentialExchangeRecord, ProofExchangeRecord } from '@aries-framework/core'
-
 import { V1RequestPresentationMessage } from '@aries-framework/anoncreds'
+import { CredentialExchangeRecord, ProofExchangeRecord, ProofState } from '@aries-framework/core'
 import { useAgent } from '@aries-framework/react-hooks'
 import { useNavigation } from '@react-navigation/core'
 import { StackNavigationProp } from '@react-navigation/stack'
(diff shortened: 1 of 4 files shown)


bryce-mcmath (commented):
Can we look at where that metadata is being set and try and fix the flow there rather than use a setTimeout?

wadeking98:
> Can we look at where that metadata is being set and try and fix the flow there rather than use a setTimeout?

Yes I've been trying that but I can't seem to resolve the issue. The code to set the metadata is being called, and I can confirm that the record is updated, but for some reason it gets reset/overwritten or maybe ignored by the notification list. I've tried looking at all places where the metadata is written on the proof record and none of them should interfere with writing `details_seen: true`. The funny thing is the same code works for setting the meta data when viewing the actual notification. If I go to the proofDetails screen via actually viewing the notification then the meta data is set correctly, but if I go to the same screen from the mobile verifier screen then the metadata is never set from the notification list's perspective. There's some weird race condition going on with the order the proofRecord is added to the notification list, and when it's metadata is updated.

I could spend some more time working on it to pin down the issue, it would probably take another work day, or I could make a ticket and come back to fixing it later.

I'll let @cvarjao and @jleach chime in too

wadeking98:
@bryce-mcmath I was able to fix it without using the setTimeout

bryce-mcmath (dismissed):
Beauty, well done

bryce-mcmath (approved):

PR #880, opened 2023-07-11
feat: restore verifier flow

wadeking98:
# Summary of Changes

becuase we didn't have goal codes set up before, when Jason implemented the changes to the connection flow, he had to remove the loading screen functionality in the mobile verifier flow. This PR restores the origional flow and adds a hook to fetch the oob record from a conneciton ID which gives you the goal code. 

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/hooks/connections.ts
+++ b/packages/legacy/core/App/hooks/connections.ts
@@ -20,7 +20,7 @@ export const useOutOfBandById = (oobId: string): OutOfBandRecord | undefined => 
       }
     })
   }
-  return useMemo(() => oob, [oobId, oob])
+  return oob
 }
 
 export const useOutOfBandByConnectionId = (connectionId: string): OutOfBandRecord | undefined => {
--- a/packages/legacy/core/App/hooks/connections.ts
+++ b/packages/legacy/core/App/hooks/connections.ts
@@ -10,23 +10,20 @@ export const useConnectionByOutOfBandId = (outOfBandId: string): ConnectionRecor
   )
 }
 
-export const useOutOfBandByConnectionId = (connectionId: string): OutOfBandRecord | undefined => {
-  const connection = useConnectionById(connectionId)
-  return useOutOfBandById(connection?.outOfBandId ?? "")
-}
-
 export const useOutOfBandById = (oobId: string): OutOfBandRecord | undefined => {
   const { agent } = useAgent()
   const [oob, setOob] = useState<OutOfBandRecord | undefined>(undefined)
   if (!oob) {
-    agent?.oob.findById(oobId).then(res => {
+    agent?.oob.findById(oobId).then((res) => {
       if (res) {
         setOob(res)
       }
     })
   }
-  return useMemo(
-    () => oob,
-    [oobId, oob]
-  )
+  return useMemo(() => oob, [oobId, oob])
+}
+
+export const useOutOfBandByConnectionId = (connectionId: string): OutOfBandRecord | undefined => {
+  const connection = useConnectionById(connectionId)
+  return useOutOfBandById(connection?.outOfBandId ?? '')
 }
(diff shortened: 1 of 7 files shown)


packages/legacy/core/App/screens/Connection.tsx
```diff
@@ -77,65 +81,81 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
...
+  useEffect(() => {
+    if (
+      connectionId &&
+      oobRecord &&
+      (!goalCode || (!goalCode.startsWith('aries.vc.verify') && !goalCode.startsWith('aries.vc.issue')))
+    ) {
       // No goal code, we don't know what to expect next,
       // navigate to the chat screen.
       navigation.navigate(Screens.Chat, { connectionId })
       dispatch({ isVisible: false })
```
jleach, line 139: Do you need a return after line 122 to stop the following lines from executing?

packages/legacy/core/App/screens/Connection.tsx
```diff
@@ -77,65 +81,81 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
...
-  //     return () => abortTimer
-  //   }, [])
-  // )
+    if (state.notificationRecord && goalCode) {
+      if (goalCode.startsWith('aries.vc.issue')) {
+        navigation.navigate(Screens.CredentialOffer, { credentialId: state.notificationRecord.id })
+      } else if (goalCode.startsWith('aries.vc.verify')) {
+        navigation.navigate(Screens.ProofRequest, { proofId: state.notificationRecord.id })
+      } else {
+        throw new Error('Unhandled notification type')
```
jleach: "Unhandled goal code type"?

packages/legacy/core/App/screens/Connection.tsx
```diff
@@ -77,65 +81,81 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
...
-    // in higher protocol versions.
-    if (connection.state === DidExchangeState.Completed || connection.state === DidExchangeState.RequestSent) {
-      dispatch({ connectionIsActive: true })
-      agent?.mediationRecipient.initiateMessagePickup()
-
+  useEffect(() => {
+    if (
+      connectionId &&
+      oobRecord &&
+      (!goalCode || (!goalCode.startsWith('aries.vc.verify') && !goalCode.startsWith('aries.vc.issue')))
```
jleach, line 134: This `if` may get gnarly the more codes we support. Maybe use a `fn` that returns a yes or now like `isSupportedGoalCode` and check an array of codes we support.

const supportedGoalCodes = ['aries.vc.verify', 'aries.vc.issue']
const isSupportedGoalCode(code: string) => supportedGoalCodes.some(c => code.startsWith(c))

jleach (dismissed):

packages/legacy/core/App/screens/Connection.tsx
wadeking98: done

packages/legacy/core/App/hooks/connections.ts
```diff
@@ -9,3 +9,21 @@ export const useConnectionByOutOfBandId = (outOfBandId: string): ConnectionRecor
...
+  const { agent } = useAgent()
+  const [oob, setOob] = useState<OutOfBandRecord | undefined>(undefined)
+  if (!oob) {
+    agent?.oob.findById(oobId).then((res) => {
+      if (res) {
+        setOob(res)
+      }
+    })
+  }
+  return useMemo(() => oob, [oobId, oob])
```
bryce-mcmath: not really sure the `useMemo` is really made use of here, and because the logic isn't in a `useEffect` I don't think it will respond to the `oobId` changing and update `oob`. (Not sure if we need that functionality anyway but still)

wadeking98: initially oob is undefined, I used UseMemo so that once they are defined then the function updates. I'm actually not sure if I need it to watch oobId, I'll update the code

bryce-mcmath (approved):

Issue #676, opened 2023-03-22
feat : Remote OCA bundle resolver


cvarjao:
Support for OCA bundle files stored remotely (via http)

Acceptance Criteria:
- Given an base URL (e.g.: `https://raw.githubusercontent.com/bcgov/aries-oca-bundles/main/OCABundles/`), the wallet should be able to discover/download/cache OCA bundle files as needed. Each wallet implementation may provide a different base URL.
- A wallet implementation (e.g.: BC Wallet) may provide a list of bundle IDs(?) to warm up cache during app initialization

```
GET /all.json
  {
    "<schema_id> or <cred_def_id>":{},
    "<schema_id> or <cred_def_id>":{},
  }

GET /<schema_id>/bundle.json
  {
     "captureBase": {},
     "overlays": [
        {"capture_base": ""}
      ]
  }

GET /<schema_id>/<cred_def_id>/bundle.json
  {
     "captureBase": {},
     "overlays": [
        {"capture_base": ""}
      ]
  }
```

thiagoromanos:
I would sugest the [GET] /all.json to return a map instead of a array, so the resolver knows what it's getting.

thiagoromanos:
Just opened a draft for the remote OCA bundle resolver and the extraction of the current bifold OCA bundle resolver.

cvarjao:
@thiagoromanos , some related conversation in bcgov/aries-oca-bundles#22 . @swcurran is suggesting an index file that is cacheable for quick lookup. Any thoughts on that?

swcurran:
Here is what the generated file is like:

https://github.com/bcgov/aries-oca-bundles/blob/main/ocabundles.json

thiagoromanos:
I think the index file is a great idea. That doesn't excludes the "/all.json" endpoint, right? (fetch each bundle at startup would be slower than fetching a single larger file)

thiagoromanos:
> Here is what the generated file is like:
> 
> https://github.com/bcgov/aries-oca-bundles/blob/main/ocabundles.json

The remote resolver is implemented and merged into bifold (#891). There is still room for upgrades, but it is able to read the index file mentioned by @swcurran and load bundles from it (just like oca explorer).

PR #883, opened 2023-07-12
feat: implement the remote OCA resolver

thiagoromanos:
# Summary of Changes

Implementing the remote OCA resolver.

# Related Issues

#676 

# Pull Request Checklist

Pro Tip 🤓


thiagoromanos:
Tagging @amanji

packages/oca/src/index.ts
```diff
@@ -2,6 +2,4 @@ export * from './formatters'
 export * from './interfaces'
 export * from './types'
 export * from './utils'
-
-// DEPRECATED: refrain from using these exports
```
amanji, line 6: We should keep this note. Eventually we will move away from legacy code.

packages/oca/src/interfaces/data/branding/BrandingOverlayData.interface.ts
```diff
@@ -4,7 +4,7 @@ export default interface IBrandingOverlayData extends IBaseOverlayData {
   logo: string
   background_image: string
   background_image_slice: string
-  primary_background_color: string
+  primary_background_color?: string
```
amanji, line 7: Any reason this is now optional or is it a carry over from something else?

packages/oca/src/legacy/resolver/oca.ts
amanji, line 1: This is already in `/legacy/types`

packages/oca/src/legacy/resolver/record.ts
amanji, line 1: This is already in `/legacy/types`

packages/oca/src/legacy/resolver/remote-oca.ts
amanji, line 1: To keep things consistent can add this to `/legacy/types`

packages/oca/src/legacy/resolver/utils/cred-def.ts
amanji, line 1: Can get this from the top level `/utls` It's using the same functions

packages/oca/src/legacy/resolver/utils/schema.ts
amanji, line 1: Can get this from the top level /utls It's using the same functions

packages/oca/src/types/branding/BrandingOverlay.ts
```diff
@@ -19,7 +19,8 @@ export default class BrandingOverlay extends BaseOverlay {
     this.logo = overlay.logo
     this.#background_image = overlay.background_image
     this.#background_image_slice = overlay.background_image_slice
-    this.#primary_background_color = overlay.primary_background_color ?? generateColor(credentialDefinitionId)
+    this.#primary_background_color =
+      overlay.primary_background_color ?? generateColor(credentialDefinitionId ?? 'default')
```
amanji: This should resolve to a valid color or an empty string. Where would `'default'` be used?

thiagoromanos: When this overlay is used on a proof. The proof screen doesn't provide a cred-def-id, so this ensures that a color will always be set

packages/oca/src/interfaces/data/branding/BrandingOverlayData.interface.ts
thiagoromanos: Because line 22 of `types/branding/BrandingOverlay.ts` . If primary_background_color is not optional, that color will never be generated based on credential-definition-id

packages/oca/src/legacy/resolver/oca.ts
thiagoromanos: The one in `/legacy/types` is not using the overlays defined in the package. Should keep the resolver folder or move the resolver to types?

packages/oca/src/legacy/resolver/record.ts
thiagoromanos: They are the same file. One of those should be removed indeed

amanji: Ok lets just remove the ones from `/types`

packages/oca/src/types/branding/BrandingOverlay.ts
amanji: Hmm. I think that logic should stay separate from this for now, a `default` is not really compatible with the specification requiring a Hex code or RGB color

packages/oca/src/interfaces/data/branding/BrandingOverlayData.interface.ts
amanji: Could we get around this by using a conditional OR instead on line 22? (i.e. `||`)

packages/oca/package.json
```diff
@@ -36,6 +36,7 @@
   "dependencies": {
     "@aries-framework/anoncreds": "^0.4.0",
     "@aries-framework/core": "0.4.0",
+    "axios": "^1.4.0",
```
amanji, line 39: Could we instead rely on the Fetch API built in? Or would this be better for consistency?

packages/oca/tsconfig.json
```diff
@@ -27,7 +25,7 @@
     // "moduleDetection": "auto",                        /* Control what method is used to detect module-format JS files. */
 
     /* Modules */
-    "module": "ESNext" /* Specify what module code is generated. */,
+    "module": "CommonJS" /* Specify what module code is generated. */,
```
amanji, line 28: I think you can keep this as "ESNext" unless it has caused issues in reuse. I think I tested it in Bifold and it didn't seem to cause an issue.

packages/oca/package.json
thiagoromanos: I used axios mainly because it gives me a "client" object that has a baseUrl, which is essential when I'm fetching relative paths. I tested the Fetch API and it worked fine, but axios was easier to handle relative paths

packages/oca/src/legacy/resolver/oca.ts
amanji: Can remove the `/types` folder in favour of this

packages/oca/tsconfig.json
thiagoromanos: I used the same that AFJ core package used. ESNext gave me some runtime errors.

packages/oca/package.json
amanji: I've also seen some implementation differences between browser Fetch API vs React Native as well.

packages/oca/src/types/branding/BrandingOverlay.ts
```diff
@@ -3,16 +3,16 @@ import { generateColor } from '../../utils/color'
 import BaseOverlay from '../base/BaseOverlay'
 
 export default class BrandingOverlay extends BaseOverlay {
-  #background_image: string
-  #background_image_slice: string
+  #background_image?: string
+  #background_image_slice?: string
   #primary_background_color: string
```
amanji, line 8: This should be optional too?

packages/oca/src/legacy/resolver/oca.ts
```diff
@@ -203,93 +144,119 @@ export class OCABundle implements OCABundleType {
     return this.getOverlay<MetaOverlay>(OverlayType.Meta10, this.options.language)
   }
 
-  public get cardLayoutOverlay(): CardLayoutOverlay10 | CardLayoutOverlay11 | undefined {
-    return this.getOverlay(this.options?.cardOverlayType || CardOverlayType.CardLayout11)
+  public get brandingOverlay(): BrandingOverlay | LegacyBrandingOverlay | undefined {
+    return this.getOverlay(this.options?.brandingOverlayType || BrandingOverlayType.Branding10)
   }
 
   public buildOverlay(name: string, language: string): MetaOverlay {
```
amanji, line 153: Not sure where this is used but seems like it's only building a `MetaOverlay` so the method name is confusing. I know it was added before this, but just making a comment. Nothing for you to change at this point

packages/oca/src/legacy/resolver/oca.ts
```diff
@@ -203,93 +144,119 @@ export class OCABundle implements OCABundleType {
...
+    } else if (metaOverlay?.issuer) {
+      colorHash = metaOverlay.issuer
     }
 
-    const cardLayoutOverlay10: CardLayoutOverlay10 = {
-      captureBase: '',
-      type: OverlayType.CardLayout10,
-      backgroundColor: hashToRGBA(hashCode(colorHash)),
+    const brandingoOverlay01: ILegacyBrandingOverlayData = {
+      capture_base: '',
```
amanji, line 239: Mixed casing may break the component. Wonder if we just keep it in camelCase for now and map over the values as needed

packages/oca/src/legacy/resolver/oca.ts
```diff
@@ -203,93 +144,119 @@ export class OCABundle implements OCABundleType {
...
+      type: OverlayType.Branding01,
+      backgroundColor: generateColor(colorHash),
     }
 
-    const cardLayoutOverlay11: CardLayoutOverlay11 = {
-      captureBase: '',
-      type: OverlayType.CardLayout11,
-      primaryBackgroundColor: hashToRGBA(hashCode(colorHash)),
+    const brandingoOverlay10: IBrandingOverlayData = {
+      capture_base: '',
```
amanji, line 245: Same comment above about mixed casing

packages/oca/src/legacy/resolver/oca.ts
amanji: So technically this wont match the interface. You may have to specify one internally for now.

packages/oca/src/types/branding/BrandingOverlay.ts
thiagoromanos: I think it should, thinking about a bundle which only defines the bg image or the slice one (or none and just sets the color for example)

MosCD3:
Please add documentation for the API changes

packages/oca/src/legacy/resolver/oca.ts
```diff
@@ -203,93 +160,129 @@ export class OCABundle implements OCABundleType {
     return this.getOverlay<MetaOverlay>(OverlayType.Meta10, this.options.language)
   }
 
-  public get cardLayoutOverlay(): CardLayoutOverlay10 | CardLayoutOverlay11 | undefined {
-    return this.getOverlay(this.options?.cardOverlayType || CardOverlayType.CardLayout11)
+  public get brandingOverlay(): BrandingOverlay | LegacyBrandingOverlay {
+    // making sure that it wont be undefined
```
thiagoromanos: I'm trying here to avoid undefined branding overlays. This can happen if a bundle is defined but the branding overlay is not defined or it is defined, but the selected branding overlay type is other than the defined one. Any thoughts on this case?

amanji (approved):

amanji:
Nice work on this!

PR #889, opened 2023-07-17
Enable references to bundles in oca-bundle.json

thiagoromanos:
# Summary of Changes

Add the feature of loading bundle references in bundles json. In oca-bundles.json asset, if the value is a string, it is a reference for another bundle. 

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/oca/src/legacy/resolver/oca.ts
+++ b/packages/oca/src/legacy/resolver/oca.ts
@@ -188,10 +188,30 @@ export class DefaultOCABundleResolver implements OCABundleResolverType {
   protected bundles: Record<string, OverlayBundle> = {}
   protected options: OCABundleResolverOptions
 
-  public constructor(bundlesData: Record<string, IOverlayBundleData> = {}, options?: OCABundleResolverOptions) {
+  public constructor(
+    bundlesData: Record<string, IOverlayBundleData | string> = {},
+    options?: OCABundleResolverOptions
+  ) {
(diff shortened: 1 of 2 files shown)


thiagoromanos:
Tagging @amanji

packages/oca/src/legacy/resolver/oca.ts
```diff
@@ -188,10 +188,30 @@ export class DefaultOCABundleResolver implements OCABundleResolverType {
...
+        }
+      } catch (error) {
+        // might get an error trying to parse javascript's default value
+      }
+    }
+
+    // getting the bundles from its key
+    for (const cid in bundlesData) {
+      try {
+        if (typeof bundlesData[cid] === 'string') {
```
amanji: Is there any chance we would encounter have some sort of recursive reference (i.e a bundle id is a reference to another reference)?

thiagoromanos: It might. I'm looking into the code, and there is another place where it checks for aliases (the `resolve` method from the legacy code). The legacy code didn't check for recursions also. It may lead to errors in casting the types in that case. Will do some refactoring to the code to avoid duplication.

amanji: I don't think it did before but we may as well add that in as an improvement

amanji (approved):

PR #891, opened 2023-07-17
feat: use @hyperledger/aries-oca package in bifold.

thiagoromanos:
# Summary of Changes

# Related Issues

# Pull Request Checklist

Pro Tip 🤓


thiagoromanos:
Waiting for the next `@hyperledger/aries-oca` npm package.

bryce-mcmath:
Still working through this, looks great so far. Can we add some additional tests so that overall coverage is not decreased?

packages/legacy/core/App/components/misc/CredentialCard10.tsx
```diff
@@ -38,7 +38,7 @@ const { width } = Dimensions.get('window')
  | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 |
  | 2 |   |   |   |   |   |   |   |
  | 3 |   |   |   |   |   |   |   |
- | 4 |   |   |   |   |   |   |   |
+ | 4 |   |   |   |   |   |   |   |1
```
amanji: Extra 1 here

amanji:
> Still working through this, looks great so far. Can we add some additional tests so that overall coverage is not decreased?

Tbh, I'm a bit confused as to how the coverage is even determined based on the report

amanji:
@thiagoromanos The util functions that have been moved over to the OCA package can be removed from Bifold IMO and referenced from the package if needed

packages/legacy/core/App/components/misc/CredentialCard10.tsx
thiagoromanos: might be a missclick while changing this file. Will revert that line.

thiagoromanos:
> @thiagoromanos The util functions that have been moved over to the OCA package can be removed from Bifold IMO and referenced from the package if needed

I think I removed those functions already. Will double check to see if I didn't missed anything, but I also think that the code that migrated to aries-oca package should be removed from bifold.

thiagoromanos:
> Still working through this, looks great so far. Can we add some additional tests so that overall coverage is not decreased?

I think we could, but don't really know where to add more tests. Most of the warnings from codecov were on lines that wasn't added, but changed, and those weren't covered already.

packages/legacy/core/App/components/misc/CredentialCard.tsx
```diff
@@ -19,7 +18,7 @@ interface CredentialCardProps {
   onPress?: GenericFn
   style?: ViewStyle
   proof?: boolean
-  displayItems?: (Attribute | Predicate)[]
+  displayItems?: (legacy.Attribute | legacy.Predicate)[]
```
amanji: I know you have already made the changes but would it be easier to reference the types at the top of the file. That way you're not having to append everything with `legacy.`

For example:

```
const Attribute = legacy.Attribute;
const Predicate = legacy.Predicate;
...
```

In the future we can still refer to `Attribute`, `Predicate`, etc. but they can be swapped out for modern components without having to replace every instance of `legacy.`

What do you think?

thiagoromanos: That's a nice idea. Future refactoring will be easier indeed.

thiagoromanos: Actually, I don't think it will work so smoothly. Some types it works fine (like `new legacy.DefaultOCABundleResolver` becoming `new DefaultOCABundleResolver`), but enums like `BrandingOverlayType` dont work that fine in parameter type definition for example 
```
  const getCredOverlayType = (type: BrandingOverlayType) => {
```
And I don't think leaving mixed use of this approach may make the code confusing

MosCD3:
Really good work @thiagoromanos my only concern here is the double work will be needed to remove that "legacy" from the code being refactored. one option is to make a simple type mapping in one place for all those "legacy" types so that ones swapped with newer version, you only change in the mapping layer. Other option is to ditch "legacy" and keep the import from top file dictate what this type comes from

packages/legacy/core/App/components/misc/CredentialCard.tsx
amanji: I was actually thinking we may want to export that Branding Overlay Type Enum directly. It can be used in the new package code as well. I know it would mean another `aries-oca` release but its much cleaner.

thiagoromanos:
> Really good work @thiagoromanos my only concern here is the double work will be needed to remove that "legacy" from the code being refactored. one option is to make a simple type mapping in one place for all those "legacy" types so that ones swapped with newer version, you only change in the mapping layer. Other option is to ditch "legacy" and keep the import from top file dictate what this type comes from

@MosCD3, I don't know how I would make the mapping for the types, but ditching the `legacy` from the type looks better IMO. However, the imports will be `@hyperledger/aries-oca/build/legacy`. I don't really like the _build_ folder in imports, but to do the refactoring later will be better.

packages/legacy/core/App/components/misc/CredentialCard.tsx
thiagoromanos: > I was actually thinking we may want to export that Branding Overlay Type Enum directly. It can be used in the new package code as well. I know it would mean another `aries-oca` release but its much cleaner.

Thats a nice idea. It will need another package release, but it's worth the effort. Will open the PR for that.

thiagoromanos: Created the PR #901 and moved the other enums that also might be useful later. As soon this one be merged (and it runs the auto publish), I'll update this PR to use the new references.

packages/legacy/core/App/components/misc/CredentialCard.tsx
```diff
@@ -1,12 +1,11 @@
 import { CredentialExchangeRecord } from '@aries-framework/core'
+import { Attribute, BrandingOverlayType, Predicate } from '@hyperledger/aries-oca/build/legacy'
```
amanji, line 2: Ok for now but eventually let's refactor this to not reference from `/build`

thiagoromanos: Thats the idea. Using `legacy.` in every reference will make it harder to refactor once the legacy tipes are gone. I don't like the `build` folder either, but since it won't be definitive, I think it's ok for now.

amanji (dismissed):

amanji:
LGTM! Not sure how we can increase the coverage again, but I'm ok merging it as is

thiagoromanos:
@bryce-mcmath , I enabled some tests in `CredentialDetails` screen and the coverage increased.

jleach:
@thiagoromanos If you can fix those two conflicts I'll merge this first thing in the morning so you don't have any more problems or delays.

jleach (approved):

packages/legacy/core/package.json
```diff
@@ -110,6 +111,7 @@
     "@formatjs/intl-relativetimeformat": "^9.2.3",
     "@hyperledger/anoncreds-react-native": "^0.1.0",
     "@hyperledger/aries-askar-react-native": "^0.1.1",
+    "@hyperledger/aries-oca": "1.0.0-alpha.72",
```
cvarjao, line 114: @thiagoromanos , I just notice this. Is this intention to use a pinned/published version as opposed to a current workspace dependency? I am not sure if it is best practice to have all workspace dependency to resolve locally from the source? or should they mix and match? looking at the yarn.lock file there seem to have 2 resolutions for `@hyperledger/aries-oca`

PR #901, opened 2023-07-24
feat: moving enums from legacy to use in new types in aries-oca package

thiagoromanos:
# Summary of Changes

Moving enums that were in legacy folder to be exported with the new types.

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/oca/src/types/TypeEnums.ts
+++ b/packages/oca/src/types/TypeEnums.ts
@@ -1,0 +1,17 @@
+export enum CaptureBaseAttributeType {
+  Binary = 'Binary',
+  Text = 'Text',
+  DateTime = 'DateTime',
+  Numeric = 'Numeric',
+  DateInt = 'DateInt',
+}
+
+export enum OverlayType {
+  CaptureBase10 = 'spec/capture_base/1.0',
+  Meta10 = 'spec/overlays/meta/1.0',
+  Label10 = 'spec/overlays/label/1.0',
+  Format10 = 'spec/overlays/format/1.0',
+  CharacterEncoding10 = 'spec/overlays/character_encoding/1.0',
+  Branding01 = 'aries/overlays/branding/0.1',
+  Branding10 = 'aries/overlays/branding/1.0',
+}
--- a/packages/oca/src/legacy/resolver/oca.ts
+++ b/packages/oca/src/legacy/resolver/oca.ts
@@ -9,6 +9,7 @@ import {
 } from '../../interfaces'
 import {
   BaseOverlay,
+  BaseOverlayType,
   BrandingOverlay,
   CaptureBase,
   CharacterEncodingOverlay,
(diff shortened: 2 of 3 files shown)


packages/oca/src/types/TypeEnums.ts
```diff
@@ -0,0 +1,20 @@
+export enum BaseType {
+  Binary = 'Binary',
+  Text = 'Text',
+  DateTime = 'DateTime',
+  Numeric = 'Numeric',
+  DateInt = 'DateInt',
+}
+
+export enum BaseOverlayType {
+  Base10 = 'spec/capture_base/1.0',
```
amanji: Small nitpick. I’d call this `CaptureBase10` since we’re renaming.

thiagoromanos: Good point. Changing it now.

packages/oca/src/types/TypeEnums.ts
```diff
@@ -0,0 +1,20 @@
+export enum BaseType {
+  Binary = 'Binary',
+  Text = 'Text',
+  DateTime = 'DateTime',
+  Numeric = 'Numeric',
+  DateInt = 'DateInt',
+}
+
+export enum BaseOverlayType {
```
amanji: I would just rename this to OverlayType we can also combine Branding into here

packages/oca/src/types/TypeEnums.ts
```diff
@@ -0,0 +1,20 @@
+export enum BaseType {
```
amanji: Not sure how relevant these really are anymore, if we have the option to remove them we should

thiagoromanos: I think they are useful to enum the types allowed on "attributes" map from "capture_base" object. If it is the case, I'll move it back to the legacy folder (bifold uses it to get the attr type in the credential)

packages/oca/src/types/TypeEnums.ts
thiagoromanos: I left it separated because they are not an "official" oca overlay. For bifold use, it makes easier to select which type of credential card was chosen. I'll make an "extra type" on legacy oca to keep it working, but combine overlay types seems to be a good idea.

packages/oca/src/types/TypeEnums.ts
amanji: Ok, not to be pedantic but lets maybe name it to `CaptureBaseType` just so it's clear it's a type in that specific object

packages/oca/src/types/TypeEnums.ts
```diff
@@ -0,0 +1,17 @@
...
+  Binary = 'Binary',
+  Text = 'Text',
+  DateTime = 'DateTime',
+  Numeric = 'Numeric',
+  DateInt = 'DateInt',
+}
+
+export enum BaseOverlayType {
```
amanji: I'd also just call this `OverlayType`

packages/oca/src/types/TypeEnums.ts
thiagoromanos: Right, gonna do that. I think `CaptureBaseAttributeType` describes it better. Any thoughts?

packages/oca/src/types/TypeEnums.ts
thiagoromanos: Good idea. Renaming it now.

amanji (approved):

PR #921, opened 2023-08-09
feat: change pin

wadeking98:
# Summary of Changes

Added functionality to change the wallet pin from the settings menu.
Updated Askar to fix rekey bug.
rekey

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/contexts/auth.tsx
```diff
@@ -132,6 +135,28 @@ export const AuthProvider: React.FC = ({ children }) => {
...
   }
 
+  const rekeyWallet = async (oldPin: string, newPin: string, useBiometry?: boolean): Promise<boolean> => {
+    const secret = await loadWalletSalt()
+    if (!secret) {
+      return false
+    }
+    const oldHash = await hashPIN(oldPin, secret.salt)
+    const newSecret = await secretForPIN(newPin)
+    const newHash = await hashPIN(newPin, newSecret.salt)
```
bryce-mcmath: These `await` statements aren't wrapped in a try catch block in case something goes wrong with them, we might want to change that

packages/legacy/core/App/screens/PINCreate.tsx
```diff
@@ -210,19 +244,46 @@ const PINCreate: React.FC<PINCreateProps> = ({ setAuthenticated }) => {
...
           <Button
-            title={t('PINCreate.CreatePIN')}
-            testID={testIdWithKey('CreatePIN')}
-            accessibilityLabel={t('PINCreate.CreatePIN')}
+            title={updatePin ? t('PINCreate.ChangePIN') : t('PINCreate.CreatePIN')}
+            testID={updatePin ? testIdWithKey('PINCreate.ChangePIN') : testIdWithKey('CreatePIN')}
+            accessibilityLabel={updatePin ? t('PINCreate.ChangePIN') : t('PINCreate.CreatePIN')}
             buttonType={ButtonType.Primary}
             disabled={!continueEnabled}
             onPress={async () => {
```
bryce-mcmath, line 265: Can we possibly move this function out of the JSX and into a variable in the component logic? Would be easier to read. Not a must, just a nice to have

bryce-mcmath (changes requested):
Couple minor comments

thiagoromanos (commented):
Aside from the `try / catch` mentioned by @bryce-mcmath, LGTM

packages/legacy/core/App/contexts/auth.tsx
wadeking98: none of those functions throw an error so a try-catch block wouldn't do anything, right?

wadeking98: I don't think awaiting something has the chance to create an error does it? await just resolves a promise but there's no timeout or anything that would cause await to throw an error, right?

packages/legacy/core/App/screens/PINCreate.tsx
wadeking98: done

packages/legacy/core/App/contexts/auth.tsx
bryce-mcmath: Functions don't have to explicitly throw errors for errors to occur. For example, the `hashPIN` function uses argon2 https://www.npmjs.com/package/argon2 and in their docs they always wrap that same code in try catch blocks

bryce-mcmath (approved):
Nice!

PR #944, opened 2023-08-29
feat: add notifications for basic messages

bryce-mcmath:
# Summary of Changes

Adds notifications for basic message records. Only shows one notification per contact with unseen basic messages.

Here's what it looks like:
message_notification

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/screens/Home.tsx
```diff
@@ -46,7 +46,9 @@ const Home: React.FC<HomeProps> = () => {
 
   const DisplayListItemType = (item: any): Element => {
     let component: Element
-    if (item.type === 'CredentialRecord') {
+    if (item.type === 'BasicMessageRecord') {
```
jleach, line 49: I think an enum would be better for this type of check.

jleach (approved):
One comment about using an enum, not a blocker. I'd also consider checking out the "satisfies" operator in TS. It may be better used in place of "as" because it does not imply you know better than the type checking mechanics.

PR #957, opened 2023-09-05
fix: PIN validation passing with incorrect PIN

jp0707:
# Summary of Changes

Fix PIN validation passing with incorrect PIN.

# Related Issues

https://github.com/hyperledger/aries-mobile-agent-react-native/issues/912

# Pull Request Checklist

### DIFF
--- a/packages/legacy/core/App/components/inputs/PINInput.tsx
+++ b/packages/legacy/core/App/components/inputs/PINInput.tsx
@@ -26,13 +26,13 @@ const PINInput: React.FC<PINInputProps & React.RefAttributes<TextInput>> = forwa
     const { t } = useTranslation()
     const { TextTheme, PINInputTheme } = useTheme()
     const cellHeight = 48
-    const myOnChangeText = (value: string) => {
+    const onChangeText = (value: string) => {
       onPINChanged && onPINChanged(value)
       setPIN(value)
     }
     const [props, getCellOnLayoutHandler] = useClearByFocusCell({
       value: PIN,
-      setValue: myOnChangeText,
+      setValue: onChangeText,
     })
 
     const style = StyleSheet.create({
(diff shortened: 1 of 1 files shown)


packages/legacy/core/App/components/inputs/PINInput.tsx
```diff
@@ -23,13 +23,17 @@ const PINInput: React.FC<PINInputProps & React.RefAttributes<TextInput>> = forwa
...
     const [PIN, setPIN] = useState('')
     const [showPIN, setShowPIN] = useState(false)
-    const [props, getCellOnLayoutHandler] = useClearByFocusCell({
-      value: PIN,
-      setValue: setPIN,
-    })
     const { t } = useTranslation()
     const { TextTheme, PINInputTheme } = useTheme()
     const cellHeight = 48
+    const myOnChangeText = (value: string) => {
```
bryce-mcmath: Looks good but can we name this function `onChangeText` and pass it directly to the `CodeField` prop of the same name?

packages/legacy/core/App/components/inputs/PINInput.tsx
```diff
@@ -80,8 +84,7 @@ const PINInput: React.FC<PINInputProps & React.RefAttributes<TextInput>> = forwa
             value={PIN}
             rootStyle={style.codeFieldRoot}
             onChangeText={(value: string) => {
```
bryce-mcmath: As per above comment:

`onChangeText={onChangeText}`

bryce-mcmath (changes requested):
One small change, otherwise good catch!

packages/legacy/core/App/components/inputs/PINInput.tsx
jp0707: done

packages/legacy/core/App/components/inputs/PINInput.tsx
jp0707: done

bryce-mcmath (approved):

PR #980, opened 2023-09-26
feat: option to rename contacts

bryce-mcmath:
# Summary of Changes

Enables users to rename their contacts (locally). I opted to use `TouchableOpacity` for now because `Pressable` was missing too much in the way of indication that it was pressed. I think after the RN upgrade we can go through and change all of our `TouchableOpacity` components to `Pressable`s at the same time and add something uniform to indicate pressed state. But in the mean time the buttons in the Contact Details screen really needed some indication that they were indeed buttons.

Here's what it looks like:
contact_rename

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/screens/Chat.tsx
```diff
@@ -54,12 +54,28 @@ const Chat: React.FC<ChatProps> = ({ route }) => {
...
   const proofs = useProofsByConnectionId(connectionId)
-  const theirLabel = useMemo(() => connection?.theirLabel || connection?.id || '', [connection])
+  const isFocused = useIsFocused()
   const { assertConnectedNetwork, silentAssertConnectedNetwork } = useNetwork()
   const [messages, setMessages] = useState<Array<ExtendedChatMessage>>([])
   const [showActionSlider, setShowActionSlider] = useState(false)
   const { ChatTheme: theme, Assets } = useTheme()
   const { ColorPallet } = useTheme()
+  const [theirLabel, setTheirLabel] = useState(
+    (connection?.id && store.preferences.alternateContactNames[connection.id]) ||
```
jleach: This check is done several times, here, `ContactDetails.tsx`, `ProofDetails.tsx`, and maybe others. Wondering if it should be simplified to a fn? Maybe something similar in `ProofRequest.tsx`?

packages/legacy/core/App/screens/ContactDetails.tsx
```diff
@@ -79,25 +87,55 @@ const ContactDetails: React.FC<ContactDetailsProps> = ({ route }) => {
...
-      })}
+      <View style={styles.contentContainer}>
+        <Text style={{ ...TextTheme.headingThree }}>{contactLabel}</Text>
+        <Text style={{ ...TextTheme.normal, marginTop: 20 }}>
+          {t('ContactDetails.DateOfConnection', {
+            date: connection?.createdAt ? formatTime(connection.createdAt, { includeHour: true }) : '',
+          })}
+        </Text>
+      </View>
+      <TouchableOpacity
```
jleach, line 116: We should decide, perhaps at a Bifold meeting, if we roll with `TouchableOpacity` or `Pressable`.

packages/legacy/core/App/screens/RenameContact.tsx
```diff
@@ -0,0 +1,161 @@
...
+type ErrorState = {
+  visible: boolean
+  title: string
+  description: string
+}
+
+type RenameContactProps = StackScreenProps<ContactStackParams, Screens.RenameContact>
+
+const RenameContact: React.FC<RenameContactProps> = ({ route }) => {
+  if (!route?.params) {
```
jleach: I wonder if this is needed? It's a pattern we do elsewhere but I think TypeScript won't let this be passed is as empty AFAIK.

packages/legacy/core/App/screens/RenameContact.tsx
```diff
@@ -0,0 +1,161 @@
...
+    throw new Error('RenameContact route params were not set properly')
+  }
+
+  const { connectionId } = route.params
+  const connection = useConnectionById(connectionId)
+  const { t } = useTranslation()
+  const { ColorPallet, TextTheme } = useTheme()
+  const navigation = useNavigation()
+  const [store, dispatch] = useStore()
+  const [contactName, setContactName] = useState(
```
jleach: For two I think `useState` is fine but `useReducer` may be more simple. It allows allows for checking for example one can check and do logic before accepting the new state. No change requested. Just a reminder its a thing.

packages/legacy/core/App/screens/RenameContact.tsx
bryce-mcmath: Will remove

packages/legacy/core/App/screens/ContactDetails.tsx
bryce-mcmath: Ok. For now I'm gonna go with `Touchable` because it has the visible reactivity that is needed by default

packages/legacy/core/App/screens/Chat.tsx
bryce-mcmath: Done, sorry that was a bit lazy of me. Added tests for the helper as well

jleach (approved):

PR #1000, opened 2023-10-18
feat: add supported languages to configuration context

fc-santos:
# Summary of Changes

Currently, all wallets need to support the three languages configured in the bifold (English, French and Portuguese). The aim of this PR is to enable you to change the default languages, so that you can choose only those you want to support.

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


jleach:
👋🏻  This can be done in Settings already - no?

 IMG_1357

fc-santos:
Hello @jleach, the goal of this PR is to only use the language you want to support in your wallet (like the bcwallet and the qcwallet) and the wanted result would look like this: 

image

It would make sense to remove Portuguese, since it's not a official language for the bcwallet and qcwallet. But to be able to do that, we need to have a new property in the ConfigurationContext so it can be overwritten in your own wallet.

thiagoromanos:
I think it's a nice feature, but as I see we are still locked to the 3 current languages. It would be awesome if we could let english as the default bifold language, and other languages could be added (using configuration like you did). So, if I want to add another language to my bifold extended wallet (spanish, for example) I could.

fc-santos:
> I think it's a nice feature, but as I see we are still locked to the 3 current languages. It would be awesome if we could let english as the default bifold language, and other languages could be added (using configuration like you did). So, if I want to add another language to my bifold extended wallet (spanish, for example) I could. 

In my opinion, it is still ok to keep any other languages that will be in the future added to the bifold. It's up to your own wallet to pick only the ones you want to support. The only modification I needed to do in my wallet was to get only the English and French translation resources as it can be seen here in this merged PR: https://github.com/MCN-ING/Portefeuille-mobile-qc/pull/64

fc-santos:
@jleach @thiagoromanos, do you guys have any other input about this PR ? Or maybe about any solutions to this problem ?

thiagoromanos:
I still think that we could use this PR to make possible to add new languages. One ideia could be implement a provider to get those language resources (the language array in `Languages.txn`, and TranslationResources and Locales enum in `localization/index.ts`. This way we could use bifolds default values, you could filter pt-BR language out from the language array in your wallet, and add a new language to some wallet would be possible too.
(That's one idea. Other ideas are welcome too)

thiagoromanos:
If new languages could not be possible to implement in this PR, I can open an issue for that. That would be a great feature to expand bifold.

fc-santos:
@jleach @thiagoromanos here are some changes done in the last commit : 

The ConfigurationContext still has a supportedLanguages prop that is now of type Locales[].

It's also possible now to pass a defaultLanguage (optional) in the initLanguage function. It was always English by default before. The languages array in the Language Screen is not hardcoded anymore and in the Settings Screen the currentLanguage doesn't use anymore the hardcoded languages array.

In the supportedLanguages prop from ConfigurationContext you can pass the locales you want to support (ex: [Locales.en, Locales.fr) or just leave the values by default([Locales.en, Locales.fr, Locales.ptBR]).

Eventually, it would be a nice feature to not display the option in the Settings -> App Settings for the changing the language if your version of the wallet supports only 1 language (for example Portuguese, which is the only official language in Brazil). So you would have only the options to Change PIN and Biometrics.

thiagoromanos (commented):
Still locked to the 3 current languages, but this refactoring will make it **way easier** to implement new language features. LGTM.

thiagoromanos:
Hey, @jleach , can you take a look at this PR too? I cannot review the PR, just put comments in it.

thiagoromanos (commented):
LGTM

wadeking98 (approved):

PR #1025, opened 2023-11-13
feat: new scan workflow

jleach:
# Summary of Changes

New scan workflow that better allows the user to understand they cannot scan different types of barcodes nor can it scan a credit card or physical ID and add it to the wallet.

# Related Issues

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/constants.ts
```diff
@@ -37,7 +40,7 @@ export const minute = 60000
 export const hour = 3600000
 
 // wallet timeout of 5 minutes, 0 means the wallet never locks due to inactivity
-export const walletTimeout = minute * 5
+export const walletTimeout = minute * 1
```
bryce-mcmath: Did you mean to commit this change?

packages/legacy/core/App/navigators/ConnectStack.tsx
```diff
@@ -17,17 +17,22 @@ const ConnectStack: React.FC = () => {
...
   return (
     <Stack.Navigator
-      // below is part of the temporary gating of the new scan screen tabs feature
-      screenOptions={{ ...defaultStackOptions, headerShown: store.preferences.useConnectionInviterCapability }}
+      screenOptions={{
+        ...defaultStackOptions,
+      }}
     >
       <Stack.Screen name={Screens.Scan} component={scan} />
+      <Stack.Screen name={Screens.ScanHelp} component={ScanHelp} />
```
bryce-mcmath, line 28: Did you mean to make this component come from config so it can be overridden? I'm assuming that's why the `whereToUseWalletUrl` in `ScanHelp.tsx` was to a throwaway url

bryce-mcmath (changes requested):
Seems to be some missing configurable-ness, but otherwise looks good. I could be misunderstanding something

packages/legacy/core/App/constants.ts
jleach: Nope. Thx.

packages/legacy/core/App/navigators/ConnectStack.tsx
jleach: For now nothing but the text needs to change and I **think** that can be set via the translation mechanics - no? I'll give it another test and make sure it can be.

bryce-mcmath: Oh ok. I thought at least the URL would have to be different for each wallet project or beyond that even the general content of the ScanHelp screen would have to be different as well

jleach: Confirmed. Translation work as expected although I am missing them in BC Wallet. Will fix here but that should not impact this PR as it is only for Bifold.

bryce-mcmath (dismissed):

bryce-mcmath (approved):

PR #1046, opened 2023-12-12
feat: updated camera to react-native-vision 3

### LINKED ISSUE #1653: Update or replace camera library
- [react-native-vision-camera](https://github.com/mrousavy/react-native-vision-camera)
- [expo camera](https://docs.expo.dev/versions/latest/sdk/camera/)

wadeking98:
# Summary of Changes

Replaced the depricated RNCamera module with react-native-vision-camera. Updated tests to be compatible. I had to make some modifications to the android permissions text... but it looks like they weren't being used anyway for some reason. This should be a transparent change, I've tested in the android emulator API version 30 and 32, as well as an iphone physical device. I was running into an issue on API version 30, but I wasn't sure if that was just a problem with the emulator itself. Other than android API 30 it works great on all versions. I'd appreciate some help testing the changes on different versions especially on android.

Additionally I've made sure the torch functionality works well.

#### Other advantages of this library:
It provides support for non QR code barcodes out of the box so we can provide more specific error messages if the user tries to scan their driver's license, etc.

# Related Issues

Fixes bcgov/bc-wallet-mobile#1653

# Pull Request Checklist

Pro Tip 🤓


bryce-mcmath:
What was the issue on API 30? Just so we can look out for it

bryce-mcmath:
Looks like you may need to update the gradle lockfile in Bifold too

wadeking98:
> What was the issue on API 30? Just so we can look out for it

I'd get a repetitive error saying failed to load depricated vision dynamite module

packages/legacy/app/package.json
```diff
@@ -81,6 +81,7 @@
     "react-native-toast-message": "^2.1.6",
     "react-native-uuid": "^2.0.1",
     "react-native-vector-icons": "^10.0.0",
+    "react-native-vision-camera": "^3.6.16",
```
jleach, line 84: Able to remove the old `"react-native-camera": "^3.31.1"` package?

packages/legacy/core/package.json
```diff
@@ -208,6 +209,7 @@
     "react-native-toast-message": "^2.1.6",
     "react-native-uuid": "^2.0.1",
     "react-native-vector-icons": "^10.0.0",
+    "react-native-vision-camera": "*",
```
jleach, line 212: Same. Ax the old `"react-native-camera": "^3.31.1"`?

jleach (changes requested):

jleach:
@wadeking98 We go back pretty far for Android API compatibility - no? Do you think we'll loose anything with this change?

wadeking98:
I've fixed the issue for api level 29 but I'm still running into issues with API level 30: https://github.com/mrousavy/react-native-vision-camera/issues/2287

wadeking98:
I've confirmed this is working with API level 29, 30, 31 now

packages/legacy/core/App/constants.ts
```diff
@@ -1,5 +1,14 @@
...
+import StaticSafeAreaInsets from 'react-native-static-safe-area-insets'
+
 import { PINValidationRules } from './types/security'
 
+export const SCREEN_WIDTH = Dimensions.get('window').width
+export const SCREEN_HEIGHT = Platform.select<number>({
+  android: Dimensions.get('screen').height - StaticSafeAreaInsets.safeAreaInsetsBottom,
+  ios: Dimensions.get('window').height,
+}) as number
+
```
bryce-mcmath: How does this behave on tablets when they switch between portrait and landscape on the camera screen?

wadeking98: good question, BC Wallet is locked in portrait mode so it shouldn't be an issue

bryce-mcmath: I don't think it's locked in portrait mode on tablets, only on phones

wadeking98: I tested it on the pixel tablet with auto rotate turned on and the BC Wallet app was locked in portrait mode on that device

bryce-mcmath: Hmm. Maybe it's just unlocked on iPads. We should ask @jleach

wadeking98: @bryce-mcmath it works fine on Ipads in either orientation, the only issue is that in landscape mode moving the camera is inverted, so left is down, right is up, etc. Trying to see if I can fix that issue now

wadeking98: Seems like this is a known issue with the library:
- https://github.com/mrousavy/react-native-vision-camera/issues/1455
- https://github.com/mrousavy/react-native-vision-camera/issues/1891

wadeking98: I could downgrade to version 2 or switch to another library but that may introduce it's own dependency issues and IMHO that seems like a lot of work to address what would be a minor inconvenience to only a handful of users, especially when it looks like this issue is going to be fixed in the library eventually. What are your thoughts @bryce-mcmath @jleach ?

bryce-mcmath: I would tend to agree with that but what happened with testing older Android versions that we still support? I remember you saying something like API 24 wasn't scanning

wadeking98: Actually I just tested it again on my mac environment and it worked fine on API 24. I think maybe I just didn't try moving the camera around last time

bryce-mcmath: Well then I'm ok to move forward with this. Might want to wait until Clecio gets back and takes a look at it but I think you've done a great job

packages/legacy/app/App.tsx
```diff
@@ -42,6 +44,10 @@ const App = () => {
     SplashScreen.hide()
   }, [])
 
+  if (!isTablet()) {
+    Orientation.lockToPortrait()
```
jleach, line 48: I think we want to [support](https://github.com/bcgov/bc-wallet-mobile/issues/1274) landscape but just lock the camera screen to portrait.

packages/legacy/app/App.tsx
```diff
@@ -42,6 +44,10 @@ const App = () => {
     SplashScreen.hide()
   }, [])
 
+  if (!isTablet()) {
```
jleach, line 47: There are some helper functions that use the screen dimensions for this. I think its better if we don't use both. If not in this PR can you create a technical debt ticket to clean this up at a later point?

packages/legacy/core/App/constants.ts
```diff
@@ -1,5 +1,14 @@
+import { Dimensions, Platform } from 'react-native'
+import StaticSafeAreaInsets from 'react-native-static-safe-area-insets'
+
 import { PINValidationRules } from './types/security'
 
+export const SCREEN_WIDTH = Dimensions.get('window').width
+export const SCREEN_HEIGHT = Platform.select<number>({
```
jleach: Why not `import { useWindowDimensions } from 'react-native'`? It should be reactive and update on-demand?

jleach (dismissed):

packages/legacy/app/App.tsx
wadeking98: I thought we decided against landscape mode in mobile view?

bryce-mcmath (changes requested):
Can we only use the official package `isTablet` function and get rid of our helper version of it?

jleach (dismissed):

packages/legacy/core/App/components/misc/ScanCamera.tsx
```diff
@@ -12,13 +12,13 @@ interface Props {
...
+  const [orientation, setOrientation] = useState(OrientationType.PORTRAIT)
   const [cameraActive, setCameraActive] = useState(true)
   const orientationDegrees: { [key: string]: string } = {
-    PORTRAIT: '0deg',
-    'LANDSCAPE-LEFT': '270deg',
-    'PORTRAIT-UPSIDEDOWN': '180deg',
-    'LANDSCAPE-RIGHT': '90deg',
+    [OrientationType.PORTRAIT]: '0deg',
+    [OrientationType['LANDSCAPE-LEFT']]: '270deg',
+    [OrientationType['PORTRAIT-UPSIDEDOWN']]: '180deg',
```
bryce-mcmath, line 20: Is one of these duplicated and we're missing landscape left now? Or am I confused

bryce-mcmath (changes requested):

packages/legacy/core/App/components/misc/ScanCamera.tsx
wadeking98: all 4 orientations are there. I don't see what you're referring to with duplicate/missing entries

wadeking98: right I accidentally duplicated portrait upside-down when I was updating things. Fixed now

bryce-mcmath (approved):

jleach (approved):

bryce-mcmath (approved):

PR #1053, opened 2023-12-31
fix: iOS app permissions

MosCD3:
# Summary of Changes

iOS 14+ uses local network permissions at app start which triggers permissions package to start, incorrectly configured Podfile and outdated package version cause an exception trigger at iOS app startup
Updated `react-native-permissions` package and added configuration in Podfile
Also tested Android to make sure updating package doesn't affect the flow

IMG_0272

# Pull Request Checklist


cvarjao:
@MosCD3 , can't we use the [reactNativePermissionsIOS.json ](https://github.com/hyperledger/aries-mobile-agent-react-native/blob/main/packages/legacy/app/reactNativePermissionsIOS.json) configuration file instead of changing the cocoa pod configuration file?

jleach:
This may be unique to your development environment. I don't see this. I do some times see, when the app starts, it ask for permission to use the network. I just answer yes and it seems to work fine. I think this is proper behaviour for iOS though.

MosCD3:
keep it a bit until I retest if not a solution I will close it

jleach:
@MosCD3 I was running Bifold the other day and ran into this problem. I think it may be needed here. Are you able to fix the conflicts so it can be merged?

MosCD3:
Yes sure, will do that

jleach (approved):

PR #1078, opened 2024-02-01
fix: issue with remote logging package

jleach:
# Summary of Changes

Fix issue with remote logging. The URL package that comes with metro has limited functionality so I need another way to extract the username and password from the base URL.

# Related Issues

n/a

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/remote-logs/src/transports.ts
+++ b/packages/remote-logs/src/transports.ts
@@ -56,30 +56,25 @@ export const lokiTransport: transportFunctionType = (props: LokiTransportProps) 
     ],
   }
 
-  // const [credentials, href] = lokiUrl.split('@')
-  // const [username, password] = credentials.split('//')[1].split(':')
-  // const protocol = credentials.split('//')[0]
-  // const authHeader = `Basic ${Buffer.from(`${username}:${password}`).toString('base64')}`
-
-  // console.log('Sending to Loki at:', `${protocol}//${href}`)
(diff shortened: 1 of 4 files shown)


packages/remote-logs/src/transports.ts
```diff
@@ -58,29 +59,22 @@ export const lokiTransport: transportFunctionType = (props: LokiTransportProps)
...
-    axiosConfig.url = `${protocol}//${href}`
-  }
-
-  axios(axiosConfig)
+  axios
+    .post(`${protocol}//${href}`, payload, config)
     .then((res) => {
       if (res.status !== 204) {
-        throw new Error(`Expected Loki to return 204, received ${res.status}`)
+        console.warn(`Expected Loki to return 204, received ${res.status}`)
```
cvarjao, line 74: Did you intend to leave `console.warn()` instead of exception?

jleach: If I leave a throw/exception in here then you need to wrap log in a try/catch which seems heavy handed. If you don't you end up with "Unhandled promise rejection". I could do nothing but then one would never know there is a problem and since this is the logger I can't really use logger to log an error 😕

cvarjao: That is fine. Just wanted to to double check it was intentional. 👍

bryce-mcmath (approved):

PR #1079, opened 2024-02-01
feat: Custom screen options dictionary: Onboarding stack only

MosCD3:
# Summary of Changes

Adding more customization to the wallet by specifying a per-screen options dictionary as optional input to the wallet config with default values implemented

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/app/ios/Podfile.lock
```diff
@@ -794,14 +795,15 @@ SPEC CHECKSUMS:
...
   RNSVG: d7d7bc8229af3842c9cfc3a723c815a52cdd1105
   RNVectorIcons: 8b5bb0fa61d54cd2020af4f24a51841ce365c7e9
   SocketRocket: f32cd54efbe0f095c4d7594881e52619cfe80b17
+  VisionCamera: b236f548d6507923857dade542c7e0a148ed3492
   Yoga: 86fed2e4d425ee4c6eab3813ba1791101ee153c6
 
 PODFILE CHECKSUM: 702646d9563d22bf09e336926a5fb0e2d3239409
 
-COCOAPODS: 1.13.0
+COCOAPODS: 1.11.2
```
bryce-mcmath: Able to keep the higher version of Cocoapods currently being used?

bryce-mcmath (changes requested):
One issue, also some conflicts to be fixed. Otherwise looks good

packages/legacy/app/ios/Podfile.lock
MosCD3: removed the file from PR

packages/legacy/core/App/navigators/defaultStackOptions.tsx
```diff
@@ -4,7 +4,36 @@ import { useTranslation } from 'react-i18next'
...
+    headerTintColor: OnboardingTheme.headerTintColor,
+    headerLeft: () => false,
+  },
+  [Screens.UseBiometry]: {
+    headerTintColor: OnboardingTheme.headerTintColor,
+    headerLeft: () => false,
+  },
+  [Screens.Developer]: {
+    headerTintColor: OnboardingTheme.headerTintColor,
+    headerShown: false,
```
bryce-mcmath, line 41: Sorry I just caught this now, this will change the existing behavior. Could you remove this `headerShown: false`?

bryce-mcmath (changes requested):
Tiny change requested

packages/legacy/core/App/navigators/defaultStackOptions.tsx
MosCD3: Current behaviour is 
RootStack line 288 -> headerShown -> false
Then line 363, Screens.Developer -> no value to override the flag. That means default value for this screen is headerShown = false

bryce-mcmath: It's easy to miss but on line 363 `options={{ ...defaultStackOptions, title: t('Screens.Developer'), headerBackTestID: testIdWithKey('Back') }}` so that spreading of `defaultStackOptions` includes a `headerShown: true`

jleach:
@MosCD3 Able to fix the conflicts and we'll re-review. Ty.

jleach:
@MosCD3 If this one is still a thing are you able to fix the conflicts. Ty.

bryce-mcmath:
The entire onboarding stack is now overridable with dependency injection / IoC. Would that be a better approach for this change rather than having config options for each stack and screen as well? @MosCD3 @jleach let me know what you think

MosCD3:
The  idea is to allow host wallets to customize Bifold to a pretty much granular level without having to duplicate code. in many cases a host wallet might want to customize title bar background color or font weight without having to create the whole onboarding stack and re-inject it

bryce-mcmath:
Ok, let's fix the conflicts and we'll get this merged today

bryce-mcmath (approved):

PR #1121, opened 2024-03-25
fix: double pin screen

wadeking98:
# Summary of Changes

- Added fix to stop PIN screen from appearing twice.
- Moved async storage calls to dependency injection called once in rootstack
- Promisified async storage calls to run in parallel, seeping up app startup

https://github.com/openwallet-foundation/bifold-wallet/assets/36937407/0755dd36-0ac3-4941-9e10-dff36421818a

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/container-impl.ts
```diff
@@ -43,12 +54,48 @@ export class MainContainer implements Container {
...
+      let tours = defaultState.tours
+      let onboarding = defaultState.onboarding
+
+      await Promise.all([
+        loadLoginAttempt().then((data) => {
+          if (data) {
+            loginAttempt = data
+          }
+        }),
+        loadState<PreferencesState>(LocalStorageKeys.Preferences, (val) => (preferences = val)),
```
wadeking98, line 78: I had to use the `(val) => (thing = val)` pattern here because I needed to mutate the `thing` variable inside the generic function but I couldn't use `useRef` or `useState` because react would complain about hooks being used in the wrong place. There's probably a better way to do this, feel free to mention it if you think of a better alternative

packages/legacy/core/App/screens/Splash.tsx
```diff
@@ -129,105 +110,74 @@ const Splash: React.FC = () => {
...
-              dispatch({ type: DispatchAction.DID_SEE_PREFACE })
-            }
+          // if they previously completed onboarding before preface was enabled, mark seen
+          if (!store.onboarding.didSeePreface) {
+            dispatch({ type: DispatchAction.DID_SEE_PREFACE })
+          }
 
-            if (!attemptData?.lockoutDate) {
+          if (!store.loginAttempt.lockoutDate) {
+            setTimeout(() => {
```
bryce-mcmath: What is this set timeout for?

wadeking98: old artifact, I just realized it was still there and removed it

packages/legacy/core/App/navigators/RootStack.tsx
```diff
@@ -89,6 +90,12 @@ const RootStack: React.FC = () => {
     }
   }
 
+  useEffect(() => {
+    loadState(dispatch).then(() => {
+      dispatch({ type: DispatchAction.STATE_LOADED })
+    })
+  }, [])
```
bryce-mcmath: Could there be some error handling here?

bryce-mcmath (changes requested):
Just one small thing, otherwise I think this is a great change

bryce-mcmath (approved):

Issue #1019, opened 2023-11-01
Deeplink returns error when app times out (ios and android)


CarolineLCa:
### Description of problem
Deeplink returns error when app times out after 5 minutes of inactivity.

### Expected behavior
It is expected that the connection via deeplink happens even if there is a timeout.

### Steps to reproduce
1. Leave the app open in background for 5 minutes
2. After that, access it using a valid deeplink
3.Enter the password and the error message will appear

The problem needs a solution for biometrics too.

### Screenshots and/or log output
![Screenshot_1698694760](https://github.com/hyperledger/aries-mobile-agent-react-native/assets/97122568/7d9a4a0e-8520-4045-a6fb-3c0292ec5ccc)

### Environment
- [X] Occurs on Android
- [X] Occurs on iOS

jleach:
Testing blocked by Lerna issue.

jleach:
Fixed in 1.0.17 1744 (tested on iOS).

PR #1160, opened 2024-06-11
fix: fixed error during deeplink on wallet timeout

wadeking98:
# Summary of Changes

There was a race condition while handling deeplinks after wallet lockout. When handling the deeplink the user would be directed to the application. The app would be in the unlocked state while the rootstack was checking if it should lockout the user if the app has been in the background long enough. Since the app was unlocked and has an active deeplink it would try to process the deeplink, however, if the user is about to be locked out then the agent would be deactivated in the middle of handling the deeplink, which would cause the error.

To fix this issue I added a check to test if the app has recently been in the background, if it has then it will check if the application is about to be locked out. Only then will it begin to process the deeplink

# Related Issues

https://github.com/openwallet-foundation/bifold-wallet/issues/1019

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/navigators/RootStack.tsx
+++ b/packages/legacy/core/App/navigators/RootStack.tsx
@@ -125,7 +125,19 @@ const RootStack: React.FC = () => {
       })
     }
 
-    if (agent && state.deepLink.activeDeepLink && state.authentication.didAuthenticate) {
+    if (agent && agent.isInitialized && state.deepLink.activeDeepLink && state.authentication.didAuthenticate) {
+      const currentTime = Date.now()
+      //user clicked on deeplink while app was in background
+      if (
+        (appStateVisible.match(/inactive|background/) || prevAppStateVisible.match(/inactive|background/)) &&
+        !state.preferences.preventAutoLock &&
+        walletTimeout &&
+        backgroundTime &&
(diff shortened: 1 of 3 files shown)


packages/legacy/core/App/navigators/RootStack.tsx
```diff
@@ -125,7 +125,19 @@ const RootStack: React.FC = () => {
       })
     }
 
-    if (agent && state.deepLink.activeDeepLink && state.authentication.didAuthenticate) {
+    if (agent && agent.isInitialized && state.deepLink.activeDeepLink && state.authentication.didAuthenticate) {
```
jleach: Able to add a little more description to this check. It may difficult to remember why this check is here 6mo down the road or by a new dev.

jleach (changes requested):

packages/legacy/core/App/navigators/RootStack.tsx
wadeking98: all good, I actually did some refactoring which allowed me to remove the check

bryce-mcmath (dismissed):
Clean

bryce-mcmath (approved):

jleach (approved):

PR #1163, opened 2024-06-13
fix: add logging and fix malformed deeplink handling

bryce-mcmath:
# Summary of Changes

There was an unnecessary second attempt at fetching from a scanned / deeplinked URI as if it was a redirect URL. The built in Credo methods already attempt to do that [here.](https://github.com/openwallet-foundation/credo-ts/blob/a648af507c05110f6d15f6737e2c750247d11f61/packages/core/src/utils/parseInvitation.ts#L159-L160) This change removes that second attempt and adds more logging to help troubleshoot future issues. It also makes the sole difference in handling between scan and deeplink that deeplinks will fail silently (but still log) if there are no valid query params available in the URI whereas scanning will throw and show an invalid QR code error. 

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/utils/helpers.ts
+++ b/packages/legacy/core/App/utils/helpers.ts
@@ -13,6 +13,7 @@ import {
 import {
   Agent,
   BasicMessageRecord,
+  BaseLogger,
   ConnectionRecord,
   CredentialExchangeRecord,
   CredentialState,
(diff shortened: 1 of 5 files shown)


packages/legacy/core/App/utils/helpers.ts
```diff
@@ -996,6 +998,7 @@ export const connectFromScanOrDeeplink = async (
 
   // Try built in Credo methods first
   try {
+    agent.config.logger.info(`Attempting to connect from ${isDeepLink ? 'deep link' : 'scan'}, value: ${value}`)
```
jleach: I think this should be resolved from DI now. Replace any other occurrences in the same file while you're at it. Examples in the code exists for copy.

jleach (changes requested):

jleach (dismissed):

packages/legacy/core/App/utils/helpers.ts
bryce-mcmath: done

jleach (approved):

PR #1173, opened 2024-06-17
feat: added support for pasting and sharing urls

wadeking98:
# Summary of Changes

This new change allows you to share your connection or proof urls over social media. another wallet user can then copy and paste the url in their scan screen to connect. Note I don't go through the full proof request flow because I'm just demoing on one device so by the time I leave the proof screen and go to paste the url, the proof doesn't exist anymore.

https://github.com/openwallet-foundation/bifold-wallet/assets/36937407/c2b848af-065c-4694-8671-ea9c221bda1b

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/screens/PasteUrl.tsx
+++ b/packages/legacy/core/App/screens/PasteUrl.tsx
@@ -1,0 +1,139 @@
+import { useAgent } from '@credo-ts/react-hooks'
+import { StackScreenProps } from '@react-navigation/stack'
+import React, { useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { StyleSheet, Text, View, TextInput, ScrollView, TouchableOpacity, Modal } from 'react-native'
+import { SafeAreaView } from 'react-native-safe-area-context'
+
+import Button, { ButtonType } from '../components/buttons/Button'
+import InfoBox, { InfoBoxType } from '../components/misc/InfoBox'
+import { TOKENS, useContainer } from '../container-api'
+import { useConfiguration } from '../contexts/configuration'
+import { useTheme } from '../contexts/theme'
+import { ConnectStackParams } from '../types/navigators'
+import { connectFromScanOrDeepLink } from '../utils/helpers'
+import { testIdWithKey } from '../utils/testable'
+
+export type PasteProps = StackScreenProps<ConnectStackParams>
+
+const PasteUrl: React.FC<PasteProps> = ({ navigation }) => {
+  const { ColorPallet, TextTheme } = useTheme()
+  const [pastedContent, setPastedContent] = useState<string>('')
+  const [errorMessage, setErrorMessage] = useState<{ title: string; message: string } | undefined>()
+  const { t } = useTranslation()
+  const { agent } = useAgent()
+  const container = useContainer()
+  const logger = container.resolve(TOKENS.UTIL_LOGGER)
+  const { enableImplicitInvitations, enableReuseConnections } = useConfiguration()
+  const styles = StyleSheet.create({
+    container: {
+      flex: 1,
+      backgroundColor: ColorPallet.brand.primaryBackground,
+    },
+    content: {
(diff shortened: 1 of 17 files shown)


packages/legacy/core/App/screens/PasteUrl.tsx
```diff
@@ -0,0 +1,139 @@
...
+        pastedContent,
+        agent,
+        logger,
+        navigation?.getParent(),
+        false, // isDeepLink
+        enableImplicitInvitations,
+        enableReuseConnections
+      )
+    } catch (err: unknown) {
+      setErrorMessage({ title: t('PasteUrl.ErrorInvalidUrl'), message: t('PasteUrl.ErrorInvalidUrlDescription') })
```
bryce-mcmath, line 65: Should we include `err?.message` in here anywhere to provide more context as to why it's failing?

packages/legacy/core/App/screens/PasteUrl.tsx
```diff
@@ -0,0 +1,139 @@
...
+    }
+  }
+
+  return (
+    <SafeAreaView style={styles.container}>
+      <ScrollView>
+        <Modal visible={!!errorMessage} testID={testIdWithKey('ErrorModal')} animationType="fade" transparent>
```
bryce-mcmath, line 72: A `DismissiblePopupModal` might be good here, it's scrollable incase the error is long

bryce-mcmath (changes requested):
A couple small things that might improve the error messaging, but overall a very cool feature and well done PR. You fixed a couple extra things while you were at it too, nice going

packages/legacy/core/App/screens/PasteUrl.tsx
```diff
@@ -0,0 +1,139 @@
...
+            testID={testIdWithKey('PastedUrl')}
+            style={styles.textBox}
+            numberOfLines={15}
+            multiline
+            value={pastedContent}
+            onChangeText={(text) => setPastedContent(text)}
+          />
+        </View>
+        <View style={styles.buttonContainer}>
+          <TouchableOpacity
```
jleach, line 103: Why add a `Button` inside a `TouchableOpacity`?

wadeking98: the wireframes want a popup to appear when the user clicks on the disabled continue button with an empty text box. I had to add a TouchableOpacity in order to detect a button press when the button is disabled

packages/legacy/core/App/screens/PasteUrl.tsx
wadeking98: a little bit difficult, there's multiple different errors that can result from an invalid url. It could be already scanned, it could contain incorrect parameters, invalid url scheme, invalid json, invalid invitation object, etc. All of these create a different error message. it's difficult to accurately detect what is failing due to a malformed url, and what is failing due to some other unknown error and if we should print the error.

packages/legacy/core/App/screens/PasteUrl.tsx
wadeking98: it's a bit hacky I admit, but I haven't had to make a disabled button respond to anything before, there might be a way to do it with a <View> instead but that's essentially the same as a touchable opacity just without a bit of styling

packages/legacy/core/App/screens/PasteUrl.tsx
wadeking98: Is there any way to make it look like an error popup instead of an info one? I'm not seeing anything in the component to support it

bryce-mcmath: Not currently, and if we're not including that `err.message` from above then it won't be variable length (won't need scroll), and so the InfoBox works as is. Approving

bryce-mcmath (approved):

PR #1179, opened 2024-06-25
fix: fixed how we handle deeplinks that don't have connection params

wadeking98:
# Summary of Changes

Previously when handling deeplinks coming from the BCSC app flow, the wallet would try to process the deeplinks because they didn't end like `bcwallet://` or `didcom://`. It seems the BCSC app now uses deeplinks in the form `bcwallet://bcsc/v1?did=null&status=cancel` I've changed the way we handle deeplinks in the root stack to only try to process deeplinkks if they contain connection parameters.

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/navigators/RootStack.tsx
+++ b/packages/legacy/core/App/navigators/RootStack.tsx
@@ -100,7 +100,7 @@ const RootStack: React.FC = () => {
   useEffect(() => {
     async function handleDeepLink(deepLink: string) {
       // If it's just the general link with no params, set link inactive and do nothing
-      if (deepLink.endsWith('//')) {
+      if (deepLink.search(/oob=|c_i=|d_m=|url=/) < 0) {
         dispatch({
           type: DispatchAction.ACTIVE_DEEP_LINK,
           payload: [undefined],


jleach (dismissed):

packages/legacy/core/App/navigators/RootStack.tsx
```diff
@@ -100,7 +100,7 @@ const RootStack: React.FC = () => {
   useEffect(() => {
     async function handleDeepLink(deepLink: string) {
       // If it's just the general link with no params, set link inactive and do nothing
-      if (deepLink.endsWith('//')) {
+      if (deepLink.search(/oob=|c_i=|d_m=/) < 0) {
```
bryce-mcmath: should also check for `_oob=` and `_url=`

bryce-mcmath (changes requested):

packages/legacy/core/App/navigators/RootStack.tsx
wadeking98: sounds good, it already checks for _oob= since `oob=` is inside it but I'll add url= too

bryce-mcmath (approved):

PR #1186, opened 2024-07-08
feat: add OCA checkForUpdates to the splash screen

timbl-ont:
# Summary of Changes

Adds support to check for OCA remote files at app startup - in the splash screen. This PR does not add support for OCA in the reference app, although that was tested. The reason for exclusion is that the method to publish OCA files is not formalized and the current support is experimental and only currently used by BCGov.

The reason for the PR is to allow upstream wallets to experiment with OCA.

The source code changes were mostly copied from the BCGov wallet.

# Related Issues

N/A

# Pull Request Checklist

- [ x] All commits contain a DCO `Signed-off-by` line (we use the [DCO GitHub app](https://github.com/apps/dco) to enforce this);
- [ x] Updated LICENSE-3RD-PARTY.md for any added dependencies or vendored components;
- [x ] Updated documentation as needed for changed code and new or modified features;
- [ x] Added sufficient [tests](../__tests__/) so that overall code coverage is not reduced.

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/screens/Splash.tsx
+++ b/packages/legacy/core/App/screens/Splash.tsx
@@ -1,6 +1,7 @@
 import { Agent, HttpOutboundTransport, WsOutboundTransport } from '@credo-ts/core'
 import { useAgent } from '@credo-ts/react-hooks'
 import { agentDependencies } from '@credo-ts/react-native'
+import { RemoteOCABundleResolver } from '@hyperledger/aries-oca/build/legacy'
 import { useNavigation } from '@react-navigation/core'
 import { CommonActions } from '@react-navigation/native'
 import React, { useEffect, useState } from 'react'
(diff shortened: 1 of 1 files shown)


packages/legacy/core/App/screens/Splash.tsx
```diff
@@ -210,6 +212,10 @@ const Splash: React.FC = () => {
       try {
         const credentials = await getWalletCredentials()
 
+        if (typeof ocaBundleResolver['checkForUpdates'] != "undefined") {
+          await ocaBundleResolver.checkForUpdates()
+        }
```
bryce-mcmath: Would it be cleaner to change this to just `await ocaBundleResolver.checkForUpdates?.()` (no if statement)

bryce-mcmath (commented):
Like the PR! Thanks for adding this. Just one small suggestion

packages/legacy/core/App/screens/Splash.tsx
timbl-ont: Updated without the if statement

bryce-mcmath (dismissed):

packages/legacy/core/App/screens/Splash.tsx
bryce-mcmath: Looks good! You may need to rebase your branch, it seems to be out of date with the base branch.

bryce-mcmath: Ah, and it looks like you'll need to run the lint step and fix any issues that arise. `yarn lint` from the root

bryce-mcmath (approved):

PR #1194, opened 2024-07-12
refactor: cleanup connection management

jleach:
# Summary of Changes

This PR refactors the connection handling logic so that it better deals with connection based, connectionless, and goal coads for automatic workflow handling.

# Related Issues

n/a

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/screens/Connection.tsx
```diff
@@ -147,54 +121,45 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
...
+    if (!state.notificationRecord || !state.isVisible) {
       return
     }
 
-    if (
-      !connectionId &&
-      !oobRecord &&
-      !goalCode &&
-      state.notificationRecord &&
-      state.notificationRecord.state === 'request-received'
```
bryce-mcmath, line 162: Are we good getting rid of this check? I think there still might be proofs in `Done` state in notifications. Not sure though.

jleach: I think it can go. I does not match any conditions in @cvarjao hackmd doc and I can't think of any use cases. Let's ax it for now and if something shows up in 1.0.19 we'll add it back with documentation as to why. Also, useNotifications won't return any in that state will it?

packages/legacy/core/App/screens/Connection.tsx
```diff
@@ -147,54 +121,45 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
...
-    ) {
-      logger?.info('Connection: Handling connectionless proof request')
-
-      navigation.replace(Screens.ProofRequest, { proofId: state.notificationRecord.id })
+    // connectionless proof request, we don't have connectionless offers.
+    if (!connection) {
       dispatch({ isVisible: false })
+      // TODO:(jl) Was navigation.replace, but I think it should be navigate
+      navigation.navigate(Screens.ProofRequest, { proofId: state.notificationRecord.id })
```
cvarjao: I think it should be `navigation.replace()`, we don't want users to be able to navigate back from the Connection screen

packages/legacy/core/__tests__/screens/__snapshots__/Connection.test.tsx.snap
```diff
@@ -832,209 +840,22 @@ exports[`ConnectionModal Component Test goal code loading screen auto navigate 1
 />
 `;
 
-exports[`ConnectionModal Component Updates after delay 1`] = `
```
cvarjao, line 835: are we testing the 'Taking Too Long' somewhere else?

packages/legacy/core/App/screens/Connection.tsx
```diff
@@ -147,54 +121,45 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
...
+      if (goalCode === GoalCodes.proofRequestVerify || goalCode === GoalCodes.proofRequestVerifyOnce) {
+        logger?.info(`Connection: Handling ${goalCode} goal code, navigate to ProofRequest`)
+        dispatch({ isVisible: false })
+        navigation.navigate(Screens.ProofRequest, { proofId: state.notificationRecord.id })
+      } else if (goalCode === GoalCodes.credentialOffer) {
+        logger?.info(`Connection: Handling ${goalCode} goal code, navigate to CredentialOffer`)
+        dispatch({ isVisible: false })
+        navigation.navigate(Screens.CredentialOffer, { credentialId: state.notificationRecord.id })
+      } else {
+        logger?.info(`Connection: Unable to handle ${goalCode} goal code`)
```
cvarjao, line 151: is this a dead end? If an unexpected goalCode is provided, then I think it should go to Chat as well.

packages/legacy/core/App/screens/Connection.tsx
```diff
@@ -209,59 +174,28 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
...
 
     for (const notification of notifications) {
-      // no action taken, we're already processing a notification
-      if (state.notificationRecord) {
-        logger?.info('Connection: Already processing a notification')
-        break
-      }
-
       // no action taken for BasicMessageRecords
       if (notification.type === 'BasicMessageRecord') {
```
cvarjao, line 193: Should we also explicitly skip App Attestation Proof Request?

packages/legacy/core/App/screens/Connection.tsx
```diff
@@ -209,59 +174,28 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
...
-      // OOB with `goalCode` will be checked in another `useEffect`. Expected
-      // as part of a OOB.
       if (
-        goalCode &&
-        oobRecord &&
-        connection &&
-        connection.outOfBandId === oobRecord.id &&
-        connection.id === notification.connectionId
+        (connection && notification.connectionId === connection.id) ||
+        (threadId && (notification.threadId === threadId || notification.parentThreadId === threadId))
```
cvarjao: I think this needs to be reviewed. We could use [OutOfBandRecord.getTags().invitationRequestsThreadIds?.[0]](https://github.com/openwallet-foundation/credo-ts/blob/v0.5.3/packages/core/src/modules/oob/repository/OutOfBandRecord.ts#L23) (this would hard code to process only 1 attachment), or since we are already retrieving OutOfBandRecord, we can pass `OutOfBandRecord.id` as a route param (instead of `connectionId`, and `threadId`)  and retrieve using `useOutOfBandById()` hook, we can then match any message (including multiples) attached to the record.

packages/legacy/core/App/screens/Connection.tsx
bryce-mcmath: I think it's possible for BC Wallet's `useNotifications` implementation to include proofs in state `Done` and in state `PresentationReceived`. It might be better for us to fix that logic in our implementation of that hook rather than in this screen though. So I'm good with merging this

bryce-mcmath (dismissed):

cvarjao (changes requested):
Some comments. Forgot to hit submit

packages/legacy/core/App/screens/Connection.tsx
jleach: Yes, but in BC only. Bifold doesn't use Attestation and I don't think I can correctly identify an Attestation proof without a known `cred_def_id` or `schema_id`.

jleach:
@cvarjao @bryce-mcmath I have to fix quite a few unit tests but I've pushed the main logic changes. I've tested with VC-Authn, ACM, Showcase scan & deep link. All work fine.

packages/legacy/core/__tests__/screens/__snapshots__/Connection.test.tsx.snap
jleach: This is that test. Are you referring to the `1`? If so that's not the delay, Jest just enumerates snapshots.

cvarjao: I just noticed that some timeTravel call was removed:
https://github.com/openwallet-foundation/bifold-wallet/pull/1194/files/291a20c52b5b334fc2b430259b4ec993b4b0372c#diff-a6cc51066ccab7c501ed7556aef0f7f5a58da96358f254d4f824a14ae73bce9eL138

I just wanted to make sure that we are still semantically testing the thing

packages/legacy/core/App/screens/Connection.tsx
cvarjao: I feel like we will need some service/function for `isAppAttestationProofRequest()` sort of thing, we can leave that for later. While I agree BC Wallet will have specific schema/creddef, having a configurable service/function might be helpful in the future.

packages/legacy/core/__tests__/screens/__snapshots__/Connection.test.tsx.snap
jleach: Yes, that's probably me just moving the logic. We still have a `timeTravel(10000)` when we check for the timeout message to appear.

packages/legacy/core/App/screens/Connection.tsx
jleach: Fixed.

bryce-mcmath (approved):

cvarjao (approved):

PR #1199, opened 2024-07-18
fix: connection no goal code

jleach:
# Summary of Changes

Fix an edge case where we need to process the state where we have a connection but not a valid goal code. In this case the user should be taken to the default Chat screen.

# Related Issues

n/a

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/screens/Connection.tsx
+++ b/packages/legacy/core/App/screens/Connection.tsx
@@ -15,7 +15,6 @@ import { DeliveryStackParams, Screens, Stacks, TabStacks } from '../types/naviga
 import { testIdWithKey } from '../utils/testable'
 
 import { useContainer, TOKENS } from './../container-api'
-import { constants } from '@credo-ts/core/build/modules/vc/data-integrity/libraries/jsonld-signatures'
 
 type ConnectionProps = StackScreenProps<DeliveryStackParams, Screens.Connection>
 
--- a/packages/legacy/core/App/screens/Connection.tsx
+++ b/packages/legacy/core/App/screens/Connection.tsx
@@ -15,6 +15,7 @@ import { DeliveryStackParams, Screens, Stacks, TabStacks } from '../types/naviga
 import { testIdWithKey } from '../utils/testable'
 
 import { useContainer, TOKENS } from './../container-api'
+import { constants } from '@credo-ts/core/build/modules/vc/data-integrity/libraries/jsonld-signatures'
 
 type ConnectionProps = StackScreenProps<DeliveryStackParams, Screens.Connection>
 
@@ -121,7 +122,7 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
   }, [state.shouldShowDelayMessage])
 
   useEffect(() => {
-    if (!state.isVisible) {
+    if (!oobRecord || !state.isVisible) {
       return
     }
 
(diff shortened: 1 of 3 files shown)


packages/legacy/core/App/screens/Connection.tsx
```diff
@@ -121,54 +121,78 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
...
-          })
-        )
-      }
-    } else {
-      logger?.info('Connection: Handling connection without goal code, navigate to Chat')
+    if (goalCode === GoalCodes.proofRequestVerify || goalCode === GoalCodes.proofRequestVerifyOnce) {
+      logger?.info(`Connection: Handling ${goalCode} goal code, navigate to ProofRequest`)
 
-      navigation.getParent()?.dispatch(
-        CommonActions.reset({
```
bryce-mcmath, line 165: Why no longer a reset here? I thought we wanted to avoid back navigation to the connection screen

packages/legacy/core/App/screens/Connection.tsx
```diff
@@ -121,54 +121,78 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
...
+      navigation.navigate(Screens.ProofRequest, { proofId: state.notificationRecord.id })
+
+      return
+    }
+
+    if (goalCode === GoalCodes.credentialOffer) {
+      logger?.info(`Connection: Handling ${goalCode} goal code, navigate to CredentialOffer`)
+
+      dispatch({ isVisible: false })
+      navigation.navigate(Screens.CredentialOffer, { credentialId: state.notificationRecord.id })
```
bryce-mcmath: Should be a reset too, no?

packages/legacy/core/App/screens/Connection.tsx
```diff
@@ -121,54 +121,78 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
...
+    // At this point we should be waiting for a notification
+    // to be processed
+    if (!state.notificationRecord) {
+      return
+    }
+
+    // Connectionless proof request, we don't have connectionless offers.
     if (!connection) {
       dispatch({ isVisible: false })
       navigation.replace(Screens.ProofRequest, { proofId: state.notificationRecord.id })
```
bryce-mcmath, line 153: Should this be a reset as well or is it a special case because it's connectionless?

bryce-mcmath (changes requested):

bryce-mcmath:
I also notice in some cases the navigation is using `getParent` and in some cases it isn't. Should be probably be consistent and verify that it works with both scan and deeplink

packages/legacy/core/App/screens/Connection.tsx
jleach: I still use reset, it just moved elsewhere in the code.

packages/legacy/core/App/screens/Connection.tsx
jleach: Fixed, I use "replace" here now.

packages/legacy/core/App/screens/Connection.tsx
jleach: This should be replace.

jleach:
> I also notice in some cases the navigation is using `getParent` and in some cases it isn't. Should be probably be consistent and verify that it works with both scan and deeplink

We use `getParent()` when navigating to "Chat" because its on a different navigation stack. Others are not so we don't need to do that. Or another way, we only do `getParent` when switching stacks because we have to go all the way back to the TabStack.

bryce-mcmath (approved):

PR #1214, opened 2024-07-28
feat: Implement base work to enable wallet to receive OpenID4VCI

MosCD3:
# Summary of Changes

- Add OpenID4VCI credo packages
- Add resolver module for OIDC
- Add parser utility to distinguish didcimm invitation from open-id
- Refactor scan QR code flow to re-route to OIDC flow when detecting such invitation

TODO: Implement UI and Navigation

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/modules/openid/metadata.tsx
```diff
@@ -0,0 +1,53 @@
...
+    display?: OpenId4VciCredentialSupported['display']
+    order?: OpenId4VciCredentialSupported['order']
+  }
+  issuer: {
+    display?: OpenId4VciIssuerMetadataDisplay[]
+    id: string
+  }
+}
+
+const openId4VcCredentialMetadataKey = '_paradym/openId4VcCredentialMetadata'
```
bryce-mcmath: Does this introduce a dependency on Paradym?

packages/legacy/core/App/utils/parsers.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  return false
+}
+
+export const isDidCommInvitation = (url: string) => {
+  if (url.startsWith(InvitationQrTypes.DIDCOMM)) {
+    return true
+  }
+
+  if (url.includes('c_i=') || url.includes('oob=') || url.includes('oobUrl=') || url.includes('d_m=')) {
```
bryce-mcmath, line 66: should this not account for `c_i`, `d_m`, `oob`, `_oob`, and `_url`? As it currently checks for those during scan handling

bryce-mcmath (changes requested):
Couple things

packages/legacy/core/App/modules/openid/metadata.tsx
MosCD3: No, but you are right, I should introduce a different key

packages/legacy/core/App/utils/parsers.tsx
MosCD3: Hmm, I will take a closer look

cvarjao: @MosCD3 , there was some work done to normalize and remove duplication on how URLs/Payloads are being parsed. We had some duplication and inconsistency between scanning QR Code vs open with deep link.
This is the main entry point:
https://github.com/openwallet-foundation/bifold-wallet/blob/ea2300f5186539d23cc401b90b92cb859f24a9e6/packages/legacy/core/App/utils/helpers.ts#L1000

MosCD3: Just to clarify, yes this is the entry point and I need a way to differ open-id invitation from that of did-comm
And the way I did it is through the entry point you mentioned @cvarjao to minimize code impact. I added a parser and the way the parser works is scanning the invitation URL for known did-comm tags in line 66, so the answer @bryce-mcmath is yes we need this to return the type of invitation so that we route the invitation to the correct flow. If you have any other better and more clean flow to have the same result, please suggest it here and I will refactor

packages/legacy/core/App/modules/openid/metadata.tsx
MosCD3: done

packages/legacy/core/App/utils/parsers.tsx
cvarjao: I think we should just worry about "isOpenIdPresentationRequest()" and then add the logic to handle that, and fallback to what we already have, I would hope that in the future Credo can actually parse the invitation payload and make that decision.

packages/legacy/core/App/utils/helpers.ts
```diff
@@ -1012,16 +1013,27 @@ export const connectFromScanOrDeepLink = async (
...
+
+      if (receivedInvitation?.id) {
+        navigation.navigate(Stacks.ConnectionStack as any, {
+          screen: Screens.Connection,
+          params: { oobRecordId: receivedInvitation.id },
+        })
+      }
+    } else if (invitationData.type === 'openid-credential-offer') {
+      //TODO: Impliment Navigation to display credential
+       // const record = await receiveCredentialFromOpenId4VciOffer({ agent: agent, uri: uri })
```
cvarjao: We should avoid silent errors/dead-end, if it is not supported/implemented yet, it should throw an error.

packages/legacy/core/App/utils/parsers.tsx
MosCD3: this is a nice plan

packages/legacy/core/App/utils/helpers.ts
MosCD3: Will throw an error. This should be the next PR after implementing the base work

packages/legacy/core/App/utils/parsers.tsx
MosCD3: resolved

packages/legacy/core/App/utils/helpers.ts
MosCD3: resolved

bryce-mcmath (approved):

PR #1216, opened 2024-07-30
chore: update and normalize eslint configuration

cvarjao:
# Summary of Changes

# Related Issues

# Pull Request Checklist

Pro Tip 🤓


package.json
```diff
@@ -5,7 +5,8 @@
   "workspaces": [
     "packages/*",
     "packages/legacy/core",
-    "packages/legacy/app"
+    "packages/legacy/app",
+    "packages/react-native-attestation/example"
```
jleach: I don't think we want this added. It's just some janky code that came along with the skeleton / templating when I created the package. I should either remove it or make it an actual thing.

packages/react-native-attestation/example/package.json
```diff
@@ -18,10 +19,13 @@
     "@babel/runtime": "^7.20.0",
     "@react-native/eslint-config": "^0.72.2",
     "@react-native/metro-config": "^0.72.11",
-    "metro-react-native-babel-preset": "0.76.8",
-    "babel-plugin-module-resolver": "^5.0.0"
+    "babel-plugin-module-resolver": "^5.0.0",
```
jleach: Wondering out loud why we use `bable`. Though `tsc` can do everything it can do.

eslint.config.mjs
```diff
@@ -0,0 +1,85 @@
+import globals from 'globals'
```
jleach, line 1: Why `.mjs` format?

package.json
cvarjao: if not needed, then we should delete the example folder. It is out of date and the name of the package is incorrect. I kept because from linting perspective, it was throwing an error

packages/react-native-attestation/example/package.json
cvarjao: This is a bob thing, it build typescript definition using `tsc` but the code transpiling is actually done using `babel`

eslint.config.mjs
cvarjao: It is just the new commended format that got auto-generated:
https://eslint.org/docs/v8.x/use/configure/configuration-files-new

it looks like more, and more libraries are defaulting to ECMAScript modules (ESM)

package.json
jleach: Ok. I removed it in my PR to fix the publishing. You can remove it here also.

jleach (approved):

PR #1236, opened 2024-08-19
feat: Implement UI for displaying OpenID4VCI credential offer

MosCD3:
# Summary of Changes

Implement UI for Credential offer of type OpenID4Vci

- Route to OpenID Cred. offer screen when scanning a correspondent QR code that offers Cred. of type W3C or SDJWT crew
- Implement UI for offered card and attributes 
- Implement Accept/Decline credential

# Related Issues

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/navigators/TabStack.tsx
```diff
@@ -21,7 +21,7 @@ import HomeStack from './HomeStack'
 const TabStack: React.FC = () => {
   const { fontScale } = useWindowDimensions()
   const [{ useNotifications }] = useServices([TOKENS.NOTIFICATIONS])
-  const notifications = useNotifications()
+  const notifications = useNotifications({})
```
bryce-mcmath, line 24: The way useNotifications is being used could probably be refactored here. The way it is currently implemented in this PR, notifications on the home screen won't be seen for OpenID creds.

packages/legacy/core/App/container-api.ts
```diff
@@ -141,7 +136,7 @@ export type TokenMapping = {
   [TOKENS.LOAD_STATE]: LoadStateFn
   [TOKENS.COMP_BUTTON]: Button
   [TOKENS.NOTIFICATIONS]: {
-    useNotifications: () => Array<BasicMessageRecord | CredentialRecord | ProofExchangeRecord | CustomNotification>
+    useNotifications: ({ openIDUri } : NotificationsInputProps) => NotificationReturnType
```
bryce-mcmath: I'm not sure this is the best approach with the way `openIDUri` is being passed around through so many components, and used in the notifications hook only sometimes. The hook should get all active notifications not just the ones for a certain connection uri.

packages/legacy/core/App/hooks/notifications.ts
```diff
@@ -64,5 +74,11 @@ export const useNotifications = (): Array<BasicMessageRecord | CredentialRecord
     setNotifications(notif)
   }, [basicMessages, credsReceived, proofsDone, proofsRequested, offers, credsDone])
 
+  useEffect(()=> {
+    if(openIDCredsRecieved) {
+      setNotifications([openIDCredsRecieved])
+    } 
+  }, [openIDCredsRecieved])
```
bryce-mcmath: This will clear all other notifications after the user scans an OpenID offer, no?

bryce-mcmath (changes requested):
Some thoughts. I can do some investigation into alternate approaches that don't change the `useNotifications` hook so drastically and don't require passing the uri around quite so much if you want.

packages/legacy/core/App/navigators/TabStack.tsx
MosCD3: The is a subsequent PR to support displaying OPenID in the home notifications and in credentials List, it not part of this PR. We discussed this approach with Cleo and Jason in our Friday dev calls and we agreed to pass an empty object for now since react hooks doesn't support undefined

packages/legacy/core/App/container-api.ts
MosCD3: This is not passing the data all over the place, this is defining that this function accepts a parameter of this name and type and its optional parameter. I think this is normal to have a function with optional parameters but if you have another better suggestion, please let me know. 
Considering that OpenID is totally different than DidComm, you cannot use a hooks to listen to credential state and trigger a notification. Its one way flow, with that in mind, to support OpenID in the wallet, we either have to clone every component starting from scanning the QR code all the way to the UI of accepting or verifying the credentials. I avoided that because I didn't want to crowd the wallet with duplicated components and I kept it only for the UI that uses OCA which have to be cloned

packages/legacy/core/App/hooks/notifications.ts
MosCD3: good point, will refactor to merge instead of replace

MosCD3: done

bryce-mcmath (approved):

PR #1248, opened 2024-08-28
fix: exhaustive deps lint errors

bryce-mcmath:
# Summary of Changes

Fixed the remaining eslint warnings (the vast majority of which were `exhaustive-deps`) and made our rules more strict going forward. This involved a lot `useCallback` updates to prevent infinite re-renders and some minor code changes that don't change logic in a few components to use primitives rather than objects. In some cases, I think it is a code smell that we have so many dependencies in our `useEffect`s, but I didn't want to change much besides fixing the lint warnings in place. One PR at a time. The worst offender of this is `Splash.tsx` with 18(!) dependencies in one case.

There is one snapshot change, but just because I added more data to the test messages in `Chat.test.tsx`

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/components/misc/NewQRView.tsx
```diff
@@ -130,53 +130,50 @@ const NewQRView: React.FC<Props> = ({ defaultToConnect, handleCodeScan, error, e
...
-      }}
-      icon="share-variant"
-    />
-  )
+  }, [navigation])
 
   useEffect(() => {
-    let headerRight = invitation ? scanShareUrl : undefined
+    let headerRight = invitation ? (
+      <HeaderButton
```
jleach, line 141: Curious why these needed to go inside the `useEffect`? Seems odd to have a component defined in a `useEffect` but I can't think of any good reason not to do it - just what I'm used to I suppose.

packages/legacy/core/App/components/misc/NewQRView.tsx
```diff
@@ -130,53 +130,50 @@ const NewQRView: React.FC<Props> = ({ defaultToConnect, handleCodeScan, error, e
...
-        Share.share({ message: invitation ?? '' })
-      }}
-      icon="share-variant"
-    />
-  )
+  }, [navigation])
 
   useEffect(() => {
-    let headerRight = invitation ? scanShareUrl : undefined
+    let headerRight = invitation ? (
```
jleach, line 140: Wonder if we can throw a space in this fn to make it a little easier to read. // wall-of-text.

jleach (changes requested):

packages/legacy/core/App/components/misc/NewQRView.tsx
bryce-mcmath: If they are just in the component, they will get redefined every time it renders. We are only using them when setting the navigation options when the tab changes, so we watch for that and only change them then. Maybe not the best way to do it, but better than it was.

packages/legacy/core/App/components/misc/NewQRView.tsx
bryce-mcmath: done

jleach (approved):

Issue #1251, opened 2024-09-04
Feature: In-line error messages on PIN screens


MalikJumani-OPS:
### Description of feature / user story
Currently, the wallet displays pop-up error messages on PIN screens. We propose showing in-line error messages to the users to help them easily understand the error and correct it instead of dismissing the pop up and then fixing the error.

### Acceptance Criteria
Given when the user is setting up the PIN or entering a PIN previously set
And the criteria for the PIN field hasn't been satisfied
Then the user sees error messages above the component

AND
When the user fulfills the criteria
Then the error message automatically goes away
And the user is allowed to save changes or move forward to the next step in the app

### Wireframes or relevant image assets / links
- Attached image is just for reference purposes.
- The text will need to be customizable to enable the developers to use this on components wherever applicable.
![image](https://github.com/user-attachments/assets/1f555bbd-6819-47c5-8e27-ebe56e69b65d)

laalauren1:
@MalikJumani-OPS  Has your issue already been escalated?

MalikJumani-OPS:
@laalauren1 not yet

laalauren1:
@MalikJumani-OPS can you give more details what app are you trying to set up or trying to logging in

MalikJumani-OPS:
@laalauren1 this is a feature request. I have updated the title.

laalauren1:
@ okay what app is that ?

PR #1253, opened 2024-09-09
chore: display error messages for PIN screens inline mode

MosCD3:
# Summary of Changes

Customize the wallet to allow inline style of displaying error/warning messages on Input PIN screens
This will be a non-default behaviour through an injectable

Note for reviewers
Removed the `App.tsx` file in `packages/legacy/app` since its not being used and it might be confusing for developers

A video demonstration here

https://github.com/user-attachments/assets/7d305428-90a1-497d-a479-c76edb90b290

# Related Issues

Related to Issue: #1251 

# Pull Request Checklist

Pro Tip 🤓


knguyenBC:
Can the inline error message be configurable so that it can also appear below the text input? BC's design system specifies that the error/help text is below:
image

jleach:
@MosCD3 Will revisit this PR once he is finished with the OpenID4VC PR.

packages/legacy/core/App/container-api.ts
```diff
@@ -19,12 +19,24 @@ import { NotificationReturnType, NotificationsInputProps } from './hooks/notific
...
+  Above,
+  Below,
+}
+
+export type InlineErrorConfig = {
+  enabled: boolean
+  position?: InlineErrorPosition
+  style?: ViewStyle
+}
+
```
bryce-mcmath: It might be better for these to be in `types.ts` or `InlineErrorText.tsx` to keep things close to what they are most related to and keep `container-api.ts` focused. I know there are a couple existing type definitions in this file already but I think that was a misstep on our part, I can move those in a future PR.

bryce-mcmath (dismissed):
Just one small organizational nitpick. As a side note, did you find it to be a hassle keeping the modal style errors as an option? I wonder if it would be better to just go all in on inline errors.

MosCD3:
> Just one small organizational nitpick. As a side note, did you find it to be a hassle keeping the modal style errors as an option? I wonder if it would be better to just go all in on inline errors.

I believe inline errors is the norm and should be the default. I believe we should use react-forms and some validation library like zod.

packages/legacy/core/App/index.ts
bryce-mcmath, line 1: Can we also export `InlineErrorPosition`?

packages/legacy/app/ios/ariesbifold.xcodeproj/project.pbxproj
bryce-mcmath, line 1: Changes to this file should not be committed

packages/legacy/app/ios/Podfile.lock
bryce-mcmath, line 1: This file should not be changed, use `yarn ios:setup` to pod install with the locked dependencies. We can update our pod/gem dependencies as part of separate intentional PR

bryce-mcmath (dismissed):
Few tiny things

packages/legacy/app/ios/Podfile.lock
MosCD3: yarn ios:setup is not working on my Mac
I get 
`activesupport-7.0.8 requires ruby version >= 2.7.0, which is incompatible with the current version, ruby 2.6.10p210`
An I did all tricks to update ruby version with no hope. It tries to install using openssl1.1 which is deprecated by homebrew. 
Any way, I removed the lock files since they are redundant any way

bryce-mcmath (approved):

bryce-mcmath:
@MosCD3 approved, just needs some conflicts fixed before it can be merged

PR #1258, opened 2024-09-13
fix: manage agent state

jleach:
# Summary of Changes

Shutting down and re-creating the agent in the Splash team is heavy handed and unnecessary. It may also be causing an inconsitent state in the Credo Agent resulting it it not picking up messages. This PR changes the logic so that when the app is backgrounded it stops message pickup and closes the wallet. Then, when the application is once again brought into the foreground it just needs to unlock the wallet and resume message pickup. This speeds up subsequent restarts and allows for faster initialization.

# Related Issues

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/navigators/RootStack.tsx
+++ b/packages/legacy/core/App/navigators/RootStack.tsx
@@ -72,6 +72,14 @@ const RootStack: React.FC = () => {
       // make sure agent is shutdown so wallet isn't still open
       removeSavedWalletSecret()
 
+      try {
+        await agent.wallet.close()
+
+        logger.info('Closed agent wallet')
+      } catch (error) {
+        logger.error(`Error closing agent wallet, ${error}`)
+      }
+
       dispatch({
         type: DispatchAction.DID_AUTHENTICATE,
         payload: [{ didAuthenticate: false }],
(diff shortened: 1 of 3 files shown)


packages/legacy/core/App/navigators/RootStack.tsx
```diff
@@ -159,9 +156,40 @@ const RootStack: React.FC = () => {
     state.authentication.didAuthenticate,
   ])
 
+  useEffect(() => {
+    if (!inBackground || !agent) {
+      return
+    }
+
+    agent?.mediationRecipient
```
bryce-mcmath: Since we are checking to make sure agent is defined above, is the optional chaining still needed?

packages/legacy/core/App/navigators/RootStack.tsx
```diff
@@ -159,9 +156,40 @@ const RootStack: React.FC = () => {
...
+      .stopMessagePickup()
+      .then(() => {
+        logger.info('Stopping agent message pickup')
```
bryce-mcmath: Nitpicky but shouldn't this say "Stopped agent message pickup"

packages/legacy/core/App/navigators/RootStack.tsx
```diff
@@ -159,9 +156,40 @@ const RootStack: React.FC = () => {
...
+      })
+      .catch((err) => {
+        logger.error(`Error stopping agent message pickup, ${err}`)
+      })
+
+    agent?.wallet
+      .close()
+      .then(() => {
+        logger.info('Closing agent wallet')
```
bryce-mcmath: Same as above, "Closed agent wallet"

packages/legacy/core/App/screens/Splash.tsx
```diff
@@ -230,6 +230,32 @@ const Splash: React.FC = () => {
...
+              routes: [{ name: Stacks.TabStack }],
+            })
+          )
+
+          return
+        }
+
+        logger.info('No agent initialized, creating a new one')
+
         await (ocaBundleResolver as RemoteOCABundleResolver).checkForUpdates?.()
```
bryce-mcmath: Could we move this above the early return for existing agents so that we still check for OCA updates?

packages/legacy/core/App/screens/Splash.tsx
```diff
@@ -230,6 +230,32 @@ const Splash: React.FC = () => {
...
+        if (agent) {
+          logger.info('Agent already initialized, restarting...')
+
+          try {
+            await agent.wallet.open({
+              id: walletSecret.id,
+              key: walletSecret.key,
+            })
+          } catch (error: unknown) {
+            logger.error('Error opening existing wallet', error as Error)
```
bryce-mcmath: Should we not catch here and let it become a BifoldError that gets emitted like the rest of the error handling in this func? Otherwise it may silently fail, no?

bryce-mcmath (changes requested):
Few suggestions

bryce-mcmath (approved):

PR #1266, opened 2024-09-23
feat: credential list footer component injectable

fc-santos:
# Summary of Changes

This PR adds a new option to add a ListFooterComponent to the ListCredentials Screen in the FlatList. This option would for example allow other wallets to add a button in the footer of the list. Example:

image

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/container-impl.ts
```diff
@@ -104,6 +104,7 @@ export class MainContainer implements Container {
     this._container.registerInstance(TOKENS.CONFIG, defaultConfig)
     this._container.registerInstance(TOKENS.COMPONENT_CRED_LIST_HEADER_RIGHT, () => null)
     this._container.registerInstance(TOKENS.COMPONENT_CRED_LIST_OPTIONS, () => null)
+    this._container.registerInstance(TOKENS.COMPONENT_CRED_LIST_FOOTER, () => null)
```
MosCD3, line 107: I am little worried about a potential bug here since you are defaulting the component to null. Wouldn't it be useful to do an empty component instead? 
If you tested that and the FlatList is happy then its okay

MosCD3 (changes requested):

packages/legacy/core/App/container-impl.ts
fc-santos: @MosCD3 I have tested and there was no problem being null. The ListFooterComponent prop from the FlatList can accept any of these types (including null):

image

knguyenBC:
The credential list screen also has an "add credential" button "represented by a "+" icon at the top right corner in the header. Does this PR take into account that button? I see in the screen posted above that that button is replaced with a "help" button.

fc-santos:
@knguyenBC the "add credential button" represented by a "+" icon at the top right corner can either be replaced by another component or removed. This component can be injected at the container level (by the default, it's null).

For example:

```typescript
//container-impl.ts

// By default at bifold-wallet 
this._container.registerInstance(TOKENS.COMPONENT_CRED_LIST_HEADER_RIGHT, () => null)
// In bc-wallet
this._container.registerInstance(TOKENS.COMPONENT_CRED_LIST_HEADER_RIGHT, AddCredentialButton)
// In qc-wallet
this._container.registerInstance(TOKENS.COMPONENT_CRED_LIST_HEADER_RIGHT, HelpCenterButton)
```

Nothing needs to be changed after this PR is merged in other wallets (no impact).

MosCD3 (approved):

PR #1283, opened 2024-10-10
feat: added new inactivity wrapper component

al-rosenthal:
# Summary of Changes

Created a new component: `InactivityWrapper.tsx`. This component is intended to wrap the app view to track user interactions (clicks, swipes, ect.). When an interaction is detected a timer internal to the component resets, if no interactions in a set time are made while the app is open, the user is logged out and is sent back to the pin entry screen. 

# Related Issues

[bc-wallet-mobile: 2136](https://github.com/bcgov/bc-wallet-mobile/issues/2136)

# Pull Request Checklist

Pro Tip 🤓


bryce-mcmath:
I wonder if it's worth consolidating some of the very similar code in the RootStack into this component, so both types of lockout are handled from the same place. @jleach thoughts?

bryce-mcmath:
Will also need to add this wrapper to the Bifold example app (so either in `RootStack.tsx` or in something like `packages/legacy/app/App.tsx` depending on how you implement it)

packages/legacy/core/App/components/misc/InactivityWrapper.tsx
```diff
@@ -0,0 +1,121 @@
...
+  Never = 0,
+}
+
+interface InactivityWrapperProps {
+  timeoutLength?: LockOutTime // number of minutes before timeoutAction is triggered, a value of 0 will never trigger the timeoutAction and an undefined value will default to 5 minutes
+}
+
+const InactivityWrapper: React.FC<PropsWithChildren<InactivityWrapperProps>> = ({ children, timeoutLength }) => {
+  const [logger] = useServices([TOKENS.UTIL_LOGGER])
+  useStore()
```
jleach: Any reason to call `useStore()` here?

packages/legacy/core/App/components/misc/InactivityWrapper.tsx
```diff
@@ -0,0 +1,121 @@
+import { useAgent } from '@credo-ts/react-hooks'
+import React, { PropsWithChildren, useEffect, useRef } from 'react'
+import { AppState, PanResponder, View } from 'react-native'
+import { useAuth } from '../../contexts/auth'
+import { useStore } from '../../contexts/store'
+import { DispatchAction } from '../../contexts/reducers/store'
+import { TOKENS, useServices } from '../../container-api'
+
+export enum LockOutTime {
```
jleach: Enums are somewhat of a hack in typescript. I think we favour `as const` like

```
const X = {
  Blah: 'zzzzzz',
  Blarb: 'rrrrr',
} as const
```

packages/legacy/core/App/components/misc/InactivityWrapper.tsx
```diff
@@ -0,0 +1,121 @@
...
+  OneMinute = 1,
+  ThreeMinutes = 3,
+  FiveMinutes = 1,
+  Never = 0,
+}
+
+interface InactivityWrapperProps {
```
jleach: Type or Interface? I tend to favour `type` but maybe interface is better?

Re: https://www.youtube.com/watch?v=zM9UPcIyyhQ

packages/legacy/core/App/components/misc/InactivityWrapper.tsx
```diff
@@ -0,0 +1,121 @@
...
+  timeoutLength?: LockOutTime // number of minutes before timeoutAction is triggered, a value of 0 will never trigger the timeoutAction and an undefined value will default to 5 minutes
+}
+
+const InactivityWrapper: React.FC<PropsWithChildren<InactivityWrapperProps>> = ({ children, timeoutLength }) => {
+  const [logger] = useServices([TOKENS.UTIL_LOGGER])
+  useStore()
+  const [_, dispatch] = useStore()
+  const { agent } = useAgent()
+  const { removeSavedWalletSecret } = useAuth()
+  const timeout_minutes = timeoutLength !== undefined ? timeoutLength : LockOutTime.OneMinute
```
jleach: Favour cammelCase over snake_case as a convention and descriptive variable names like `timeoutInMinues`

jleach (changes requested):

packages/legacy/core/App/components/misc/InactivityWrapper.tsx
```diff
@@ -0,0 +1,121 @@
...
+    })
+  }
+
+  const clearTimer = () => {
+    if (inactivityTimer.current) {
+      clearTimeout(inactivityTimer.current)
+    }
+  }
+
+  const minutesToMilliseconds = (minutes: number) => {
```
jleach: I might just simplify this to `const blahAsMiliseconds = minutes * 60000` at the top of the file.

packages/legacy/core/App/components/misc/InactivityWrapper.tsx
```diff
@@ -0,0 +1,121 @@
...
+        // some user interaction, reset timeout
+        resetInactivityTimeout(timeout_minutes)
+
+        // returns false so the PanResponder doesn't consume the touch event
+        return false
+      },
+    })
+  ).current
+
+  const resetInactivityTimeout = (minutes: number) => {
```
jleach: You probably need to wrap all your fn inside a `useCallback`. Every time a component renders a function will be re-created which is a performance hit. Also, because a fn can be a dependency on a `useEffect` when it is re-created it will change causing a `useEffect` to trigger. So, in general, its good to just wrap 'em and let RN sort it out.

packages/legacy/core/App/components/misc/InactivityWrapper.tsx
al-rosenthal: updated

packages/legacy/core/App/components/misc/InactivityWrapper.tsx
al-rosenthal: Happy to change that. Is there a way to still have this object act as a type when defining something like a type or interface? for example:

```typescript
const MyConst = {
  Blah: 'something'
} as const

type MyType = {
  someAttribute: MyConst
}

```

packages/legacy/core/App/components/misc/InactivityWrapper.tsx
al-rosenthal: I have no strong feelings towards either, it is a type for now.

packages/legacy/core/App/components/misc/InactivityWrapper.tsx
al-rosenthal: nope

al-rosenthal:
> Will also need to add this wrapper to the Bifold example app (so either in `RootStack.tsx` or in something like `packages/legacy/app/App.tsx` depending on how you implement it)

I've added the wrapper to the `App.tsx`

packages/legacy/core/App/components/misc/InactivityWrapper.tsx
al-rosenthal: functions are wrapped in callbacks

packages/legacy/core/App/components/misc/InactivityWrapper.tsx
al-rosenthal: good call

jleach:
@al-rosenthal You can use `yarn run lint` in `bifold` and `yarn run typecheck:core` to cache code quality issues quicker or use husky on `push`.

bryce-mcmath (commented):
This looks great but you may want to check with Kim / Clecio if an actual settings option (where the user can switch between the different lockout times) was part of the ticket.

al-rosenthal:
> This looks great but you may want to check with Kim / Clecio if an actual settings option (where the user can switch between the different lockout times) was part of the ticket.

A new settings page has been added to configure the timeout

packages/legacy/core/App/contexts/reducers/store.ts
```diff
@@ -646,6 +665,7 @@ export const reducer = <S extends State>(state: S, action: ReducerAction<Dispatc
       }
     }
     default:
+      console.log(`${action.type} is not available`)
```
bryce-mcmath: Can be helpful for local dev but shouldn't be committed, the linter will complain about `console` usage

packages/legacy/core/App/localization/en/index.ts
```diff
@@ -248,7 +248,7 @@ const translation = {
     "BiometricsNotProvided": "Biometrics not provided, you may use PIN to load the wallet.",
     "BiometricsError": "Biometrics were not successful.",
     "BiometricsErrorEnterPIN": "Please enter your wallet PIN.",
-    "LockedOut": "You locked out after 5 minutes of inactivity.",
+    "LockedOut": "You were locked out after ## minute(s) of inactivity.",
```
bryce-mcmath: There's an example of how to do this without using `String.prototype.replace` if you check the "MinutesAgo" entry.

bryce-mcmath (changes requested):
Looks really good, just two tiny changes and a rebase then I will approve

packages/legacy/core/App/contexts/reducers/store.ts
al-rosenthal: removed

packages/legacy/core/App/localization/en/index.ts
al-rosenthal: swapped out with established pattern

bryce-mcmath (approved):
Nice work!

packages/legacy/core/App/localization/fr/index.ts
```diff
@@ -654,7 +654,14 @@ const translation = {
         "Build": "Créer",
         "WhatAreContacts": "Qu’est-ce qu’un Contact?",
         "ScanMyQR": "Scanner mon code QR",
-        "Developer": "Options de développeur"
+        "Developer": "Options de développeur",
+        "AutoLockTime": "Auto lock time (fr)",
```
jleach, line 658: We tend to do this in all caps but not worth holding up the PR for.

packages/legacy/core/App/screens/PINEnter.tsx
```diff
@@ -354,7 +354,9 @@ const PINEnter: React.FC<PINEnterProps> = ({ setAuthenticated, usage = PINEntryU
     if (store.lockout.displayNotification) {
       return (
         <>
-          <Text style={style.helpText}>{t('PINEnter.LockedOut')}</Text>
+          <Text style={style.helpText}>
+            {t('PINEnter.LockedOut', { time: String(store.preferences.autoLockTime ?? 5) })}
```
jleach, line 358: I think 5 by itself is a magic number. In another PR circle back and define it somewhere like `const defautlAutolockInMinutes = 5`. Also, you have it here and in `    autoLockTime: 5, // default wallets lockout time to 5 minutes` so to change the default it needs to be updated in 2x places. Better to define it in constants and use it in both places.

packages/legacy/core/App/screens/AutoLock.tsx
```diff
@@ -0,0 +1,135 @@
...
+      borderBottomWidth: 1,
+      borderBottomColor: ColorPallet.brand.primaryBackground,
+      marginHorizontal: 25,
+    },
+    checkboxContainer: {
+      justifyContent: 'center',
+    },
+  })
+
+  const handleTimeoutChange = (time: (typeof AutoLockTime)[keyof typeof AutoLockTime]) => {
```
jleach, line 51: No need for `useCallback`? If the linter is happy then so am I.

jleach (approved):

PR #1284, opened 2024-10-10
chore: update flakey firewall check

jleach:
# Summary of Changes

The firewall check if flakey at best because of it drawing on the genesis transactions which may or may not (most likely not) represent existing nodes. Because its so hit and miss we elected just check the mediator for network connectivity. If this fails a toast is shown indicating such. If it works, nothing is displayed.

A future enhancement would be to get a connect node list form IndyVDR and test a sample of them.

# Related Issues

n/a

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/contexts/network.tsx
+++ b/packages/legacy/core/App/contexts/network.tsx
@@ -3,8 +3,8 @@ import * as React from 'react'
 import { createContext, useContext, useState } from 'react'
 
 import NetInfoModal from '../components/modals/NetInfoModal'
-import { fetchLedgerNodes, canConnectToLedgerNode } from '../utils/ledger'
-
+import { hostnameFromURL, canConnectToHost } from '../utils/network'
+import { Config } from 'react-native-config'
 export interface NetworkContext {
   silentAssertConnectedNetwork: () => boolean
   assertConnectedNetwork: () => boolean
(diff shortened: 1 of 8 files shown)


packages/legacy/core/App/contexts/network.tsx
```diff
@@ -41,13 +41,15 @@ export const NetworkProvider: React.FC<React.PropsWithChildren> = ({ children })
...
-    const nodes = fetchLedgerNodes()
+    const hostname = hostnameFromURL(Config.MEDIATOR_URL!)
 
-    if (typeof nodes === 'undefined' || nodes.length === 0) {
+    if (hostname === null || hostname.length === 0) {
       return false
     }
 
-    const connections = await Promise.all(nodes.map((n: { host: string; port: number }) => canConnectToLedgerNode(n)))
+    const nodes = [{ host: hostname, port: 443 }]
```
bryce-mcmath, line 50: Do we need this array and Promise.all and includes check if we are just checking one value?

bryce-mcmath (approved):
One comment but works for me

packages/legacy/core/App/contexts/network.tsx
jleach: No but it allows others to add hosts and they'll get magically checked also. We just happen to only care about one.

PR #1291, opened 2024-10-22
feat: add PINExplainer pre-screen to the PINCreate screen to explain the use of the PIN

jian4on:
# Summary of Changes

New screen added before PIN Set-up to explain the use of the PIN in the wallet to the user. The pre-screen is displayed on navigating to Pin Create screen, the user need to click CONTINUE to access the PIN creating form

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/screens/PINCreate.tsx
```diff
@@ -160,135 +162,146 @@ const PINCreate: React.FC<PINCreateProps> = ({ setAuthenticated, route }) => {
...
-    <KeyboardView>
-      <View style={style.screenContainer}>
-        <View style={style.contentContainer}>
-          <PINCreateHeader updatePin={updatePin} />
-          {updatePin && (
+    <>
+    { explained ||
+      <PINExplainer continueCreatePIN={continueCreatePIN} />
+    }
+    { explained &&
```
bryce-mcmath: I believe this should be a ternary expression:
```
return explained ? (<KeyboardView>...</KeyboardView>) : (<PINExplainer continueCreatePIN={continueCreatePIN} />)
```
That way we won't need the fragment and we won't risk odd UI bugs

packages/legacy/core/App/screens/PINExplainer.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+import { ScrollView, StyleSheet, Text, View } from 'react-native'
+
+import { useTheme } from '../contexts/theme'
+
+export interface PINExplainerProps {
+  continueCreatePIN: () => void
+}
+
+const Component: React.FC<PINExplainerProps> = ({ continueCreatePIN }) => {
```
bryce-mcmath:
```suggestion
const PINExplainer: React.FC<PINExplainerProps> = ({ continueCreatePIN }) => {
```
Component should be named even though it's a default export, it's the existing pattern and it helps readability.

packages/legacy/core/App/screens/PINExplainer.tsx
```diff
@@ -0,0 +1,103 @@
...
+    fill: ColorPallet.notification.infoText,
+    height: 150,
+    width: 150,
+  }
+  
+  // const navigation = useNavigation<StackNavigationProp<AuthenticateStackParams>>()
+  const pressBack = () => {
+    // No handler on Back action yet
+    // navigation.navigate(Screens.Terms)
+  }
```
bryce-mcmath: Since we're going with the component approach and it's not a separate screen, this can be removed

packages/legacy/core/App/screens/PINExplainer.tsx
```diff
@@ -0,0 +1,103 @@
+import React, { useEffect } from 'react'
+/*
+import { useNavigation } from '@react-navigation/native'
+import { StackNavigationProp } from '@react-navigation/stack'
+import { AuthenticateStackParams, Screens } from '../../types/navigators'
+*/
```
bryce-mcmath: Can be removed

packages/legacy/core/App/screens/PINExplainer.tsx
```diff
@@ -0,0 +1,103 @@
...
+          </View>
+          <View style={[{ paddingTop: 10, marginBottom: 20 }]}>
+            <Button
+              title={t('Global.Back')}
+              accessibilityLabel={t('Global.Back')}
+              testID={testIdWithKey('Back')}
+              onPress={pressBack}
+              buttonType={ButtonType.Secondary}
+            />
+          </View>
```
bryce-mcmath: Can be removed but don't forget to move the `marginBottom: 20` to the button above that we are keeping

bryce-mcmath (changes requested):
Couple small things to clean up, great first PR though!

packages/legacy/core/App/screens/PINCreate.tsx
jian4on: Agree

packages/legacy/core/App/screens/PINExplainer.tsx
jian4on: Agree

packages/legacy/core/App/screens/PINExplainer.tsx
jian4on: Agree

packages/legacy/core/App/localization/fr/index.ts
```diff
@@ -187,6 +187,16 @@ const translation = {
...
     },
     "PINCreate": {
+        "Explainer": {
+          "PrimaryHeading": "Secure your Wallet",
+          "PINReminder": "On the following screen you will be asked to set up a strong <b>personal identification number</b> (PIN).",
+          "WhyNeedPin": {
+            "Header": "Why you need a PIN",
+            "Paragraph": "Your PIN is an important security feature:",
+            "ParagraphList1": "You can use your PIN instead of biometrics to open your wallet",
+            "ParagraphList2": "Or use your biometrics to open your wallet, and if your biometrics change change on your device you will need to use the PIN to authenticate your wallet"
```
bryce-mcmath: Because this is the French translation file and these are in English, please trail each with a `(FR)` so that we know they need to be translated. You'll also need to add these to the portuguese `(PT-BR)` translation file. Sorry I didn't catch this in my first review

bryce-mcmath (changes requested):
Just one change I missed in my first review, easy fix. Sorry about that

bryce-mcmath:
Also @jian4on, tests seem to be failing. You can check those locally by running `yarn test:watch` from the `packages/legacy/core` directory. If you all you need to do is update snapshots, there will be an option to press `u` during this watch process which will take care of that.

packages/legacy/core/__tests__/screens/PINChange.test.tsx
```diff
@@ -1,4 +1,4 @@
-import { render } from '@testing-library/react-native'
+import { render, fireEvent } from '@testing-library/react-native'
```
al-rosenthal: nit: looks like this isn't getting used

packages/legacy/core/__tests__/screens/PINCreate.test.tsx
```diff
@@ -1,4 +1,4 @@
-import { render } from '@testing-library/react-native'
+import { render, fireEvent } from '@testing-library/react-native'
```
al-rosenthal: nit: looks like this isn't getting used

packages/legacy/core/App/container-api.ts
```diff
@@ -44,6 +45,7 @@ export const SCREEN_TOKENS = {
   SCREEN_SPLASH: 'screen.splash',
   SCREEN_SCAN: 'screen.scan',
   SCREEN_USE_BIOMETRY: 'screen.use-biometry',
+  SCREEN_PIN_EXPLAINER: 'component.pin-explainer'
```
al-rosenthal: should this be renamed to `screen.pin-explainer`

al-rosenthal (commented):
A few small nit picky items, otherwise looks good!

bryce-mcmath (approved):
Just needs to be rebased and then we're off 🚀

PR #1296, opened 2024-10-23
feat: OpenID4VP Proof request Base work and UI

MosCD3:
# Summary of Changes

This PR introducing the flow of accepting a proof request from an issuer using OIDC flow

# Related Issues

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/package.json
```diff
@@ -197,14 +199,14 @@
     "react-native-fs": "^2.16.6",
     "react-native-gesture-handler": "^1.10.3",
     "react-native-get-random-values": "^1.7.0",
-    "react-native-gifted-chat": "*",
```
cvarjao, line 200: I can understand that `*` is super generic, but any particular reason for setting to a very specific peer dependencies? Usually keep compatibility at major level (starting with `^`) is fine. Similar comment  for `react-native-qrcode-svg`, and `react-native-reanimated` changes below.

MosCD3: The "*" caused installing latest stable version, and it caused a whole lot of conflicts with other packages that are mot updated, and to do that in this PR is a big mess. I had to fix the version to get things aligned together and get the app to build and tests to succeed.

MosCD3: TBH, I believe leaving yarn to pickup latest updated versions for any packages is recipe for bugs. We have to dedicate PRs to solely update specific packages.

cvarjao: we can be specific at `dependencies` but more liberal with `peer dependencies`. Also, dependencies should not be updated as a side-effect, it should be explicit and managed by the lock file.

cvarjao: keep in mind that `yarn install` MAY produce unexpected dependency change. `yarn install --immutable` is the safest command to use and void unexpected changes to the lock file

MosCD3: I never meant to touch any of that :). It NEVER succeeded until I fixed versions for those packages. Let me know if you have another idea and we can hop on a call and discuss

packages/legacy/core/App/modules/openid/resolverProof.tsx
```diff
@@ -0,0 +1,279 @@
...
+        }
+      } else if (query.request && typeof query.request === 'string') {
+        const _res = {
+          data: query.request,
+          certificate: extractCertificateFromJwt(query.request),
+        }
+        return _res
+      }
+    }
+    return { data: null, certificate: null }
```
bryce-mcmath, line 143: In future, can we prefer using early returns so there isn't as many nested ifs? Like this for example:
```
if (!uri) {
  return { data: null, certificate: null }
}

// rest of conditional logic
```

packages/legacy/core/App/screens/Connection.tsx
```diff
@@ -246,6 +247,11 @@ const Connection: React.FC<ConnectionProps> = ({ navigation, route }) => {
       })
       return
     }
+
+    if ((state.notificationRecord as OpenId4VPRequestRecord).type === 'OpenId4VPRequestRecord') {
+      dispatch({ inProgress: false })
+      navigation.replace(Screens.OpenIDProofPresentation, { credential: state.notificationRecord })
```
bryce-mcmath, line 253: In future, can we avoid passing objects as navigation params? I think react-navigation recommends only using primitives as params, so we could pass a credential ID instead.

bryce-mcmath (approved):
Approved with some suggestions for future PRs

packages/legacy/core/App/screens/Connection.tsx
MosCD3: we are working in a model that will create a react-hooks for open-id work, this then will be solved

PR #1304, opened 2024-11-05
fix: small fixes

al-rosenthal:
# Summary of Changes

- Cleaned up old `App.tsx`
- added additional exports to index
- cleaned up padding

# Related Issues

[BC-Wallet 2250](https://github.com/bcgov/bc-wallet-mobile/issues/2250)

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/app/App.tsx
bryce-mcmath, line 1: I think @cvarjao recently mentioned we should keep the two `App.tsx` for now as one will eventually be an export that can be used as the default wrapper in the various Bifold projects. So this `App.tsx` will import the wrapper from core and just be an example of how to use it. Long story short, I think we keep them both for now.

bryce-mcmath (changes requested):
Just one thing, otherwise looks good

packages/legacy/app/App.tsx
al-rosenthal: restored from main

bryce-mcmath (approved):

PR #1306, opened 2024-11-07
feat: makes all header buttons injectable

ClaudeArs:
# Summary of Changes

This PR makes all header buttons injectable

# Related Issues

N/A
# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/navigators/ContactStack.tsx
```diff
@@ -16,52 +16,79 @@ import WhatAreContacts from '../screens/WhatAreContacts'
...
+          headerRight: () => <ContactListHeaderRignht />,
         })}
       />
       <Stack.Screen
         name={Screens.ProofRequest}
         component={ProofRequest}
-        options={{ title: t('Screens.ProofRequest') }}
+        options={() => ({  title: t('Screens.ProofRequest'),
+        headerRight: () => <ContactListHeaderRignht />,
+      })}
```
bryce-mcmath: One major issue I see with this PR is that these header components are the same for many different screens. So if a Bifold user wants a different header component on a different screen in the same stack or if they want no header component for a certain screen in the same stack, this PR would prevent that.

bryce-mcmath (changes requested):
One large issue. I agree it would be nice to make all the header buttons configurable, but I'm not sure this is the best way to go about it. Let's try and think of a better way. I'll spend some time thinking on it too.

packages/legacy/core/App/navigators/ContactStack.tsx
ClaudeArs: Yes I understand. I will try to find a solution to resolve this problem.
Thank you

PR #1317, opened 2024-11-15
feat: inject header right

ClaudeArs:
# Summary of Changes

This PR makes all header buttons injectable

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


bryce-mcmath (changes requested):
This is a better approach! My only two gripes are:
1) These are actually screen options, not stack options. If we call them stack options then we will have two separate and very different `defaultStackOptions` variables in our codebase. 
2) You've moved some of the screen options to the `defaultStackOptions` but not all of them. It might be good to either keep them all in place, or move them all to the new `defaultStackOptions`. I would lean towards leaving them all in place.

MosCD3:
Does have some overlap with #1321 could work out a complete solution
I am not in a favour of mixing any stackOptions with screen options

bryce-mcmath:
Ah right @ClaudeArs sorry I forgot a version of this already exists using dictionaries. See `packages/legacy/core/App/navigators/defaultStackOptions.tsx` and `OnboardingStack`. Might be good to have a chat with @MosCD3 as he was the one who implemented the first approach.

ClaudeArs:
Ok @MosCD3 , the goal of this PR is to give the possibility to inject a button in the right part of the header. According to you, what would be the best way to proceed?

ClaudeArs:
Hello @MosCD3, I allow myself to relaunch you in order to find the solution that would suit everyone best in the present case. We should find a solution in the fairly near future.

Thank you

packages/legacy/core/App/types/ScreenOptions.ts
```diff
@@ -0,0 +1,37 @@
+import { StackNavigationOptions } from "@react-navigation/stack";
+
+export interface ScreenOptions {
```
MosCD3, line 3: I think this will be a redundant re-declaration. There is already ScreenOptionsType exported, you just need to refactor to make it apply to all screens instead of OnboardingStack and rename the Token. 
That's not to mention that the naming would conflict with Global Layout options which is in this PR #1321

MosCD3 (changes requested):

packages/legacy/core/App/types/ScreenOptions.ts
ClaudeArs: Ok, I will close this PR and create a new one with the correct modifications
Thanks

ClaudeArs:
@MosCD3 I created a new PR according to your advice. [#1335](https://github.com/openwallet-foundation/bifold-wallet/pull/1335)

PR #1321, opened 2024-11-18
feat: global screen wrapper

MosCD3:
# Summary of Changes

This is a first step into having a global screen wrapper to allow more customization over pages layout

Included

- Creating the wrapper with minimal settings
- Applying to some screens as a means of initial implementation
- Applying to terms screen with flags to allow custom headers 
- Extend injectable screen options to allow injecting custom header

# Related Issues

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/types/navigators.ts
```diff
@@ -209,4 +210,9 @@ export type HistoryStackParams = {
   [Screens.HistoryPage]: undefined
 }
 
-export type ScreenOptionsType = Partial<Record<Screens, StackNavigationOptions>>
+export type StackNavigationOptionsExtended = StackNavigationOptions & {
+  screenEdges?: Edge[] | undefined
+  customHeader?: () => React.ReactNode | undefined
+}
+
+export type ScreenOptionsType = Partial<Record<Screens, StackNavigationOptionsExtended>>
```
jcdrouin21: ScreenOptionsDictionary is used in the OnboardingStack to pass options to the screen (see example below). With this change in type, the screen options can now include screenEdges and customHeader, but these two properties should not be passed to the screen options.
image

MosCD3: Yes, it shouldn't be passed to the screen options, and if you try to pass them, linting will fail. but there is no harm from using the abstract type "ScreenOptionsType" to include them for the sack of using them in the wrapper.

jcdrouin21: Yes, but the ScreenOptionsDictionary has already been passed to the screen options in the onboardingStack. The token in the useService was specifically created for this purpose, so I'm not entirely sure I understand what you're referring to.

MosCD3: It shouldn't cause any issues, but could be a concern. I will take another look at the architecture

MosCD3: done

packages/legacy/app/container-imp.tsx
```diff
@@ -20,6 +21,30 @@ export class AppContainer implements Container {
...
+        customEdges: ['bottom'],
+        safeArea: true,
+        Header: () => (
+          <View style={{ backgroundColor: 'red', height: 129, justifyContent: 'center', alignItems: 'center' }}>
+            <Text style={{ color: 'white' }}>Custom Header</Text>
+          </View>
+        ),
+      },
+    })
+      */
```
jcdrouin21: ?

packages/legacy/core/App/layout/wrapper.tsx
```diff
@@ -0,0 +1,53 @@
...
+const ScreenLayout: React.FC<PropsWithChildren<Props>> = ({ children, screen }) => {
+  //safeArea, customEdges, style, header
+  const [screenLayoutOptions] = useServices([TOKENS.OBJECT_LAYOUT_CONFIG])
+  const screenProps = screenLayoutOptions[screen]
+  const { safeArea, customEdges, style, Header } = screenProps || {
+    safeArea: false,
+    customEdges: ['top', 'left', 'right', 'bottom'],
+    style: {},
+    header: undefined,
+  }
```
jcdrouin21: It shouldn't be? :
```
const { safeArea, customEdges, style, Header } = screenProps || {
    safeArea: false,
    customEdges: ['top', 'left', 'right', 'bottom'],
    style: {},
+   Header: undefined,
  }
```

packages/legacy/core/App/navigators/defaultLayoutOptions.tsx
```diff
@@ -0,0 +1,8 @@
+import { ScreenLayoutConfig, Screens } from '../types/navigators'
+
+export const DefaultScreenLayoutOptions: ScreenLayoutConfig = {
+  [Screens.Terms]: {
+    customEdges: ['top', 'left', 'right'],
+  },
+  //TODO: Add more screens here
```
jcdrouin21, line 7: It should be clear that adding new screens here requires submitting a new PR to include the wrapper at the screen level as well. 

However, I wonder if it might be more efficient to handle everything within this PR instead.

packages/legacy/app/container-imp.tsx
MosCD3: This is an example, not sure if its a good idea to leave it here or maybe document it somewhere or remove it and developers figure out

packages/legacy/core/App/navigators/defaultLayoutOptions.tsx
MosCD3: I dont think it's a good idea to refactor all screens in this PR. Same as we did with History module, we implemented the feature and now Engineers from Quebec are picking up on it.

packages/legacy/core/App/layout/wrapper.tsx
MosCD3: Good catch, will fix it

packages/legacy/app/container-imp.tsx
al-rosenthal: I'm a little torn on this as well, what about moving this into the wrapper itself? I think it would be useful to have an example of this for someone who might not fully understand how to implement this on a screen in the future. It would also be great to have a brief explanation as to what this is an example of and maybe provide a little more context to the wrapper itself a

packages/legacy/core/App/layout/ScreenLayout.tsx
cvarjao, line 1: We might rename this to `DefaultScreenLayout` or `MainScreenLayout` over time.

packages/legacy/core/App/screens/Terms.tsx
```diff
@@ -15,6 +14,7 @@ import { useStore } from '../contexts/store'
 import { useTheme } from '../contexts/theme'
 import { AuthenticateStackParams, Screens } from '../types/navigators'
 import { testIdWithKey } from '../utils/testable'
+import ScreenLayout from '../layout/ScreenLayout'
```
cvarjao, line 17: We might rename this to `DefaultScreenLayout` or `MainScreenLayout` over time.

cvarjao (approved):

jleach:
@MosCD3 Able to up so we can merge this one in?

PR #1348, opened 2024-12-05
fix: pin explainer ui

bryce-mcmath:
# Summary of Changes

This PR makes the button on the PINExplainer screen always visible, while the content is still scrollable. I also added to the reacti18n mock so that we can test screens and components that makes use of its `Trans` component. I also fixed a jest error that was popping up when running tests, caused by the `ChatMessage` component.

pin_explainer

# Breaking change guide

N/A

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/screens/PINExplainer.tsx
+++ b/packages/legacy/core/App/screens/PINExplainer.tsx
@@ -1,10 +1,12 @@
 import React from 'react'
 import { Trans, useTranslation } from 'react-i18next'
+import { ScrollView, StyleSheet, Text, View } from 'react-native'
+import { SafeAreaView } from 'react-native-safe-area-context'
+
 import Button, { ButtonType } from '../components/buttons/Button'
 import BulletPoint from '../components/inputs/BulletPoint'
-import { testIdWithKey } from '../utils/testable'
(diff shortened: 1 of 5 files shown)


packages/legacy/core/App/screens/PINExplainer.tsx
```diff
@@ -38,8 +42,8 @@ const PINExplainer: React.FC<PINExplainerProps> = ({ continueCreatePIN }) => {
   }
 
   return (
-    <ScrollView contentContainerStyle={{ flexGrow: 1 }}>
-      <View style={style.screenContainer}>
+    <SafeAreaView style={style.safeAreaView} edges={['bottom', 'left', 'right']}>
```
al-rosenthal, line 45: is there a reason why 'top' is left out?

al-rosenthal (approved):
Looks good!

packages/legacy/core/App/screens/PINExplainer.tsx
bryce-mcmath: Because there is a header on this screen, which handles the top safe area already. Including top here would just cause extra space below the header and above the content

jleach (approved):

PR #1356, opened 2024-12-11
fix: disable back action by gesture after accepting terms and conditions

fc-santos:
# Summary of Changes

This PR fixes a bug where after accepting the Terms & Conditions the user is able to go back from the PINCreate Screen and then is stuck on Terms & Conditions Screen. Only killing app gets the user back on the PINCreate screen.

N/A

# Breaking change guide

N/A

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/navigators/defaultStackOptions.tsx
+++ b/packages/legacy/core/App/navigators/defaultStackOptions.tsx
@@ -27,6 +27,7 @@ export const DefaultScreenOptionsDictionary: ScreenOptionsType = {
   },
   [Screens.CreatePIN]: {
     headerLeft: () => false,
+    gestureEnabled: false,
   },
   [Screens.NameWallet]: {
     headerTintColor: OnboardingTheme.headerTintColor,
--- a/packages/legacy/core/App/navigators/defaultStackOptions.tsx
+++ b/packages/legacy/core/App/navigators/defaultStackOptions.tsx
@@ -26,7 +26,7 @@ export const DefaultScreenOptionsDictionary: ScreenOptionsType = {
     headerLeft: () => false,
   },
   [Screens.CreatePIN]: {
-    headerLeft: () => false,
+    //headerLeft: () => false,
     gestureEnabled: false,
   },
   [Screens.NameWallet]: {
--- a/packages/legacy/core/App/screens/PINCreate.tsx
+++ b/packages/legacy/core/App/screens/PINCreate.tsx
@@ -1,4 +1,4 @@
-import { CommonActions, ParamListBase, useNavigation } from '@react-navigation/native'
+import { CommonActions, ParamListBase, useFocusEffect, useNavigation } from '@react-navigation/native'
 import { StackNavigationProp, StackScreenProps } from '@react-navigation/stack'
 import React, { useState, useRef, useEffect, useCallback } from 'react'
 import { useTranslation } from 'react-i18next'
@@ -12,6 +12,7 @@ import {
   TouchableOpacity,
   findNodeHandle,
   DeviceEventEmitter,
+  BackHandler,
 } from 'react-native'
 import Icon from 'react-native-vector-icons/MaterialIcons'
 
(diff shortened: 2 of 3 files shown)


al-rosenthal (approved):

cvarjao:
@fc-santos , can you elaborate a bit more on the problem. Is that on iOS, Android, both? is that on Bifold or is that on your own implementation of a wallet using bifold? It seems like getting stuck on a screen by going back seems the problem.

fc-santos:
@cvarjao it's meant to fix this problem (occurs on both Android and iOS): 

https://github.com/user-attachments/assets/be93ed68-b772-46af-982a-36ca0beaf5ab

cvarjao:
I might be missing something, I don't think removing the ability to go back fix the problem, probably a workaround. It shouldn't get stuck as you are seeing. I think that is the problem that needs to be fixed

bryce-mcmath:
@fc-santos adding to what Clecio said, I suspect it has to do with the checkbox no longer existing when you come back to the Terms screen. Not having it checked is probably preventing the user from continuing. Probably better to fix that underlying bug that just preventing back nav. Might be as simple as changing line 23 of `Terms.tsx` to:

```
const [checked, setChecked] = useState<boolean>(store.onboarding.didAgreeToTerms ?? false)
```

Edit: Or you could simply remove the checkbox. We don't use one in our injected version of the Terms screen.

fc-santos:
@bryce-mcmath I agree that we should do what you suggested, but at the same time do we want to enable the user to go back though? The user has already accepted the conditions, why are allowing it to be accepted again?

fc-santos:
> @fc-santos adding to what Clecio said, I suspect it has to do with the checkbox no longer existing when you come back to the Terms screen. Not having it checked is probably preventing the user from continuing. Probably better to fix that underlying bug that just preventing back nav. Might be as simple as changing line 23 of `Terms.tsx` to:
> 
> ```
> const [checked, setChecked] = useState<boolean>(store.onboarding.didAgreeToTerms ?? false)
> ```
> 
> Edit: Or you could simply remove the checkbox. We don't use one in our injected version of the Terms screen.

@bryce-mcmath  To answer your point about removing the checkbox, it's a legal requirement for us to have it.

knguyenBC:
> @bryce-mcmath I agree that we should do what you suggested, but at the same time do we want to enable the user to go back though? The user has already accepted the conditions, why are allowing it to be accepted again?

The Pan-Can design team had a discussion on this topic. We've concluded that allowing users to go back to the terms and conditions after already accepting it it is not necessary. So removing the back gesture is not a usability or UX concern.

bryce-mcmath:
> @bryce-mcmath I agree that we should do what you suggested, but at the same time do we want to enable the user to go back though? The user has already accepted the conditions, why are allowing it to be accepted again?

@fc-santos I think generally we should prefer native-style functionality where possible as it may frustrate users if they sometimes can back nav and sometimes can't. Also, increasing the amount of custom navigation mechanics might end up as tech debt. But for this specific case, I think it's fine if we make both changes - we fix the checkbox bug, and we also prevent the back nav. That's good with me 👍

fc-santos:
@bryce-mcmath we have analyzed more the behaviour and our team agreed in the end on allowing the users to go back. I can add the check to display the checkbox so users are not stuck on the terms after moving back from the create pin screen. I’ll push the changes tomorrow.

bryce-mcmath:
> @bryce-mcmath we have analyzed more the behaviour and our team agreed in the end on allowing the users to go back. I can add the check to display the checkbox so users are not stuck on the terms after moving back from the create pin screen. I’ll push the changes tomorrow.

Sounds good! Thanks for keeping us in the loop! 🤝

fc-santos:
@bryce-mcmath this PR can be closed and point at this new PR : #1367

jleach:
Closing as per @fc-santos request. New one create #1367.

PR #1390, opened 2025-01-10
feat: Change network status check feature in NetInfo, disable/enable mediator check and provide new method to check isInternetReachable

jian4on:
# Summary of Changes
+ added Config Token to disable/enable mediator check
+ added Config Token to define internet reachability URLs, which are used as beacon for internet status testing
+ added new method in networkProvider to check isInternetReachable
+ implemented the new network check feature to NetInfo component

N/A

# Breaking change guide
N/A

# Related Issues
N/A

# Pull Request Checklist

Pro Tip 🤓


jian4on:
Will fix the __test, new provider spy failed in other feature testing suites

jian4on:
**Native Build & Test / build-ios**  failed, it is the boost installing error happens in the new year. The reason is the host changed - "_The boost download was hosted on JFrog until December 31, 2024 when the service ended. Therefore the problem appeared out of nowhere_" on Jan 1, 2025".
see the discussing at [Here](https://github.com/boostorg/boost/issues/843#issuecomment-1872943124)

packages/legacy/core/App/contexts/network.tsx
```diff
@@ -28,19 +30,34 @@ export const NetworkProvider: React.FC<React.PropsWithChildren> = ({ children })
...
     const isConnected = silentAssertConnectedNetwork()
     if (!isConnected) {
       displayNetInfoModal()
     }
-
     return isConnected
   }
 
-  const assertNetworkReachable = async (): Promise<boolean> => {
+  // Ping internet check beacon endponits, set internetReachable to be true positive response from anyone
```
timbl-ont: Should use the netinfo and not custom code for the reachability check. We can pass custom URLs to the netinfo library if required. The reason for this is that Android has built in reachability checks that netinfo will use if available. The use of google is a fallback. On IOS the netinfo library does use the fallback for now. @jian4on

timbl-ont:
Context for this PR:

When using the oid4vci flow, a mediator is not necessary. If a mediator is not present (or not needed for the wallet to function in a use case) then there needs to be another way to confirm Internet connectivity. The library react-native-netinfo is already used in the code base and is able to make use of IOS and Android functionality to test Internet reachability with lots of configurability. The other option was to create another .env variable with alternate sites to test that would be used by the existing code instead of the mediator. The assumption is that the native reachability functionality in the OS will scale better then a custom solution. Netinfo also has many small optimizations for corner cases and is well maintained.

There is still some overlap in functionality that could be consolidated. A cleaner approach in the future would be to use netinfo events to manage Internet reachability as well as specifically checking for the mediator (if configured). Note that even if the mediator is configured, a wallet may still want to exclude or manage the check differently if the mediator is not part core use case.

packages/legacy/core/App/contexts/network.tsx
```diff
@@ -28,19 +30,23 @@ export const NetworkProvider: React.FC<React.PropsWithChildren> = ({ children })
...
     if (!isConnected) {
       displayNetInfoModal()
     }
-
     return isConnected
   }
 
-  const assertNetworkReachable = async (): Promise<boolean> => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const assertInternetReachable = async (_urls?: string[]): Promise<boolean> => {
```
al-rosenthal: is there a reason to leave these parameters in?

jeznorth:
@jian4on Please reviews comments so we can proceed.

packages/legacy/core/App/contexts/network.tsx
jian4on: Fixed, it was used when multiple ping test URLs were passed to reachibility test.

al-rosenthal (approved):

PR #1401, opened 2025-01-18
feat: improve the PIN Enter repetition validation rule and add relevant messages

jian4on:
# Summary of Changes
Change the PIN repetition validation to a number type rule 'no_repeated_numbers' to make the repeated-times validation setting more readable

N/A

# Breaking change guide
The previous 'no_repeated_numbers' was an union type 'boolean | number', it is changed to a simple number. The default validation setting was '0' for allowing any times repeating.
Don't implement boolean value, the number value explanation:
0: Disable the repetition check, e.g. '111111' is allowed
1:  Enable the repetition check, '123456' is allowed but '113456' forbidden
2: Enable the repetition check, '113456' is allowed but '111456' forbidden
...

# Related Issues
N/A

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/constants.ts
```diff
@@ -71,7 +71,7 @@ export const PINRules: PINValidationRules = {
   only_numbers: true,
   min_length: 6,
   max_length: 6,
-  no_repeated_numbers: false,
```
bryce-mcmath, line 90: I know this is in draft but if it's doable could we keep the defaults the same and just allow for opt-in behavior with the injectable `PINSecurity.rules` config we have currently?

jian4on: Put both PINSecurity and PINCreationValidation to IoC container, then the wallet variants can customize them without changinf Bifold default validating behavior. I will discuss with team. Thanks @bryce-mcmath

jian4on: Back to default, but change the validating logic for '0' as 'repeating times' instead of its boolean value, keep the extra message for numbers repeating allowed but the repeating times exceeding the max times rule

packages/legacy/core/App/utils/PINCreationValidation.ts
```diff
@@ -36,16 +38,21 @@ export const PINCreationValidations = (PIN: string, PINRules: PINValidationRules
       errorName: PINError.OddOrEvenSequenceValidation,
     } as PINValidationsType)
   }
-  if (PINRules.no_repeated_numbers) {
-    let noRepeatedNumbers = new RegExp(/(\d)\1{1,}/)
-    if (typeof PINRules.no_repeated_numbers === 'number') {
-      noRepeatedNumbers = new RegExp(`(\\d)\\1{${PINRules.no_repeated_numbers - 1},}`)
-    }
+
+  if ('number' === typeof(PINRules.no_repeated_numbers) && PINRules.no_repeated_numbers > 2 ) {
```
bryce-mcmath: When would the `> 2` check be needed at the same time as the string interpolation below is? Wouldn't there only be one case (3) where that condition is satisfied?

bryce-mcmath (dismissed):
I did have one question out of curiosity but it seems to still work either way. Nice going!

packages/legacy/core/__tests__/utils/PINCreationValidation.test.ts
```diff
@@ -77,7 +77,7 @@ describe('PIN creation validations', () => {
   test('PIN with repeated numbers and repeated numbers validation to true, so the validation use the default of two repeated numbers, should return NoRepetitionOfTheSameNumbersValidation as invalid', async () => {
     const PINRulesWithRepeatedNumbers = {
       ...defaultPINRules,
-      no_repeated_numbers: true,
+      no_repeated_numbers: 0,
```
jleach: I think `no_` should be change to `num_` if it now indicates the number of of repeating values allowed.

jleach (changes requested):

packages/legacy/core/__tests__/utils/PINCreationValidation.test.ts
jian4on: @jleach Good point, It is type `'boolean | number'` now. I changed it to an `enum [1,2,3,4 ...]` type in the Draft version for discussing and then changed it back. Actually I still think maxRepetition number is an option.
@bryce-mcmath If you think we need more discussion, I will pull this feat back to draft

bryce-mcmath: @jian4on ah right, the type changed. `num` is good I think, we can change the default since the type is changing. maybe just add a note in the "Breaking changes" section of the description of this PR so other teams know what to change to keep their existing behavior

timbl-ont: Not sure this will work well. There will be no way to disable the rule unless another flag is added. You can put in a high number but that will then be displayed during password selection, which would be confusing for the user.

packages/legacy/core/App/types/security.ts
```diff
@@ -10,11 +10,16 @@ export enum AuthLevel {
...
+  0 - Disable adjacent number repeating validation, any times repeating are allowed
+  n > 0 - Enable the repeating validation, n repeating times are forbidden, e.g. n = 2 forbid repeating 2 times, '11' is allowed but '111' forbidden
+*/
 export interface PINValidationRules {
   only_numbers: boolean
   min_length: number
   max_length: number
-  no_repeated_numbers: boolean | number
+  no_repeated_numbers: number
   no_repetition_of_the_two_same_numbers: boolean | number
```
bryce-mcmath, line 23: We should probably make the same change for the rules below:
```suggestion
  no_repeated_numbers: number
  no_repetition_of_the_two_same_numbers: number
```

bryce-mcmath (approved):
One suggested change that you don't have to make, but it might make sense to do it at the same time - I'll leave it up to you. Nice work, sorry for the wait!

bryce-mcmath:
@jleach I think your re-review will be required as well

jleach (approved):

jleach:
@jian4on If you want to rebase this I'll re-review. I tried through the UI but I cannot because of a conflict.

jeznorth:
Update so we can merge this.

MosCD3 (approved):

PR #1413, opened 2025-01-24
fix: graceful shutdown patch, add agent restart

bryce-mcmath:
# Summary of Changes

This PR adds a patch to `@credo-ts/core` to fix a websocket issue that was preventing graceful shutdown. This was also addressed by the following PR in Credo:
- https://github.com/openwallet-foundation/credo-ts/pull/2152

Additionally, it adds a patch to `@credo-ts/indy-vdr` to allow agent reinitialization as per this PR:
- https://github.com/openwallet-foundation/credo-ts/pull/2137

It also adds back agent restart if shutdown was successful to cutdown on start up time after being locked out etc. and simplifies the shutdown process in Bifold

N/A

# Breaking change guide

**These patches will be required in downstream projects until we upgrade to the next Credo release (likely 0.6.x)**

Another thing that is not a breaking change per se, but if you use your own implementation of `useInitializeAgent` and want the restart functionality, you will have to add it. I would recommend doing this

# Related Issues

bcgov/bc-wallet-mobile#2243

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/hooks/initialize-agent.ts
```diff
@@ -101,6 +125,12 @@ const useInitializeAgent = () => {
...
 
+    const existingAgent = await restartExistingAgent()
+    if (existingAgent) {
+      setAgent(existingAgent)
+      return existingAgent
+    }
+
     const newAgent = await createNewAgent()
     if (!newAgent) {
       return
```
al-rosenthal, line 137: no agent to restart and it failed to create a new agent, this should probably throw an exception or at least log

packages/legacy/core/App/hooks/initialize-agent.ts
```diff
@@ -25,6 +25,30 @@ const useInitializeAgent = () => {
     TOKENS.UTIL_LEDGERS,
   ])
 
+  const restartExistingAgent = useCallback(async () => {
+    // if the agent is initialized, it was not a clean shutdown and should be replaced, not restarted
+    if (!walletSecret?.id || !walletSecret.key || !agent || agent.isInitialized) {
+      return
```
al-rosenthal, line 31: there are also a few instances of an if similar to this, it might also be good to add some logging at these returns

bryce-mcmath: This hook gets called in the splash screen which has three states: before onboarding is complete, after onboarding but before unlock, and after unlock but before agent initialization. We only want it to run if it's in that third state, and only in that third state are both `walletSecret.id` and `walletSecret.key` both defined. In the other two cases it's behaving correctly if it does nothing

al-rosenthal (approved):

PR #1420, opened 2025-01-30
feat: Add opacity overlay in Scan screen but keep the scan window transparent and clear

jian4on:
# Summary of Changes
Add opacity overlay in Scan screen upon the 'ScanCamera' zone but keep the scan window clear, so that the textual contents upon the camera video could be recognized even users scan a bright background. And the clear scan window also improve the experience for QR scan alignment

The background image in the screenshot is only for testing, it doesn't exist in real usage.

# Breaking change guide
N/A

# Related Issues
N/A

# Pull Request Checklist

Pro Tip 🤓


jian4on:
Simulator Screenshot - iPhone 16 - 2025-01-29 at 23 32 10

jian4on:
I think we should treat this file as an exception for Sonar scan rules. The centre clear box is created with View boxes nesting and some tricky styles. The code looks like duplicated but it is not.

jian4on:
The Sonar scan failed for some reason not reasonable, I think we can make an exception for this file. The clear window needs some views nesting and some tricky styles, maybe Sonar not happy with this.

packages/legacy/core/App/components/misc/QRScanner.tsx
```diff
@@ -31,30 +31,34 @@ const QRScanner: React.FC<Props> = ({ handleCodeScan, error, enableCameraOnError
...
   const { t } = useTranslation()
   const { ColorPallet, TextTheme } = useTheme()
-
   const styles = StyleSheet.create({
-    container: {
-      flex: 1,
-    },
     viewFinder: {
+      marginTop: -90,
+      borderWidth: 320,
```
bryce-mcmath, line 37: It's best to avoid using pixel values for positioning wherever possible. Pixel values can be very brittle and have unforeseen affects across different devices, causing inconsistent UIs. Could we try to make the same change using flexbox?

packages/legacy/core/App/components/misc/QRScanner.tsx
```diff
@@ -31,30 +31,34 @@ const QRScanner: React.FC<Props> = ({ handleCodeScan, error, enableCameraOnError
...
-      borderWidth: 2,
-      borderColor: ColorPallet.grayscale.white,
-    },
-    viewFinderContainer: {
-      flexGrow: 1,
-      justifyContent: 'center',
-      alignItems: 'center',
     },
     messageContainer: {
+      zIndex: 999,
```
bryce-mcmath, line 49: Changing `zIndex` should also be avoided. It affects iOS and Android differently, and the more zIndex changes in the codebase the more likely we are to run into unexpected UI appearance and styling difficulty. In this specific case, I don't see any effect of the zIndex that couldn't be achieved with just altering the order of JSX elements.

packages/legacy/core/App/components/misc/QRScanner.tsx
```diff
@@ -31,30 +31,34 @@ const QRScanner: React.FC<Props> = ({ handleCodeScan, error, enableCameraOnError
...
       marginHorizontal: 40,
       flexDirection: 'row',
       alignItems: 'center',
       alignSelf: 'center',
       paddingTop: 30,
     },
+    bottomContainer: {
+      position: 'absolute',
+      width: '100%',
+      bottom: 60,
```
bryce-mcmath, line 60: Same as first comment above

bryce-mcmath (changes requested):
Couple things to fix, but also some extra context and feedback:

The reason SonarCloud is griping about the duplication is that there is another component called `NewQRView` which is very similar but has many additional features and alterations. To see this other component, enable `Connection Inviter Capability` via developer settings and then reopen the scan screen.

The two components will eventually be merged into one and not gated behind developer mode. In the meantime we can manually dismiss the SonarCloud check on a per-PR basis, but I don't think we should add an exception into our SC config. 

The change from this PR should probably be made in both components as well

packages/legacy/core/App/components/misc/QRScanner.tsx
jian4on: There are too many things need to be discussed, especially the NewQRView component. I pull it to draft and bring the feedbacks to team, thanks.

bryce-mcmath:
@jian4on I brought this up with the rest of the BC team, and we'd like to merge the two components into one but still allow for the two variations. So the change would be making two components into one that is configured based on that connection inviter capability. In other words, there would be no change to the UI for folks using either option, but there would be less duplication. Let me know if that works for you. If it does I will tackle it ASAP so you can finish the work in this PR. 🤝

timbl-ont:
@bryce-mcmath That makes sense to do first before this PR is completed. I have not had a chance to look at the new connection inviter. Will it impact the OID4VC connection logic?

bryce-mcmath:
> @bryce-mcmath That makes sense to do first before this PR is completed. I have not had a chance to look at the new connection inviter. Will it impact the OID4VC connection logic?

It won't. Both components only affect UI on the scan screen, all the actual connection logic is passed through them by the same function (`handleCodeScan`)

bryce-mcmath:
> @bryce-mcmath That makes sense to do first before this PR is completed. I have not had a chance to look at the new connection inviter. Will it impact the OID4VC connection logic?

@timbl-ont It won't. Both components only affect UI on the scan screen, all the actual connection logic is passed through them by the same function (`handleCodeScan`)

bryce-mcmath:
@jian4on inspired by this PR, I added a new PR here that merges the two components and adds the same overlay: https://github.com/openwallet-foundation/bifold-wallet/pull/1440

check it out when you have a moment :)

PR #1433, opened 2025-02-07
feat: make attemptLockout rules injectable in the Config interface

fc-santos:
# Summary of Changes

This PR makes the attempt lockout rules injectable in the Config interface. The default values are as seen in the `constants.ts` file. If a wallet decides to make their own rules, the attemptLockoutBaseRules keys need to be multiples of the attemptLockoutThresholdRules.attemptIncrement.

N/A

# Breaking change guide

N/A

# Related Issues

This also fixes an issue where after the first 5 tries, the user gets locked out for 1 minute. After the user has served the 1 minute penalty, they can try again. But then if they make the wrong PIN, they get locked out right away and the penalty doesn't increase (it stays locked for 1 minute).

# Pull Request Checklist

Pro Tip 🤓


bryce-mcmath (commented):
Not gonna stop this PR from being merged as-is, but I wonder if these two config objects (`attemptLockoutBaseRules` and `attemptLockoutThresholdRules`) could be merged and simplified a bit - just `attemptLockoutRules`? It would be easier to read and remove the need for some of the destructuring and extra effect dependencies in `PINEnter.tsx`. What do you think?

al-rosenthal (approved):

PR #1435, opened 2025-02-07
fix: decline proof

jleach:
# Summary of Changes

This pull request fixes a bug that occurred when declining a proof request from a notification. If the contact associated with that proof request had their connection deleted, the app would throw a visible error.  This happened because the app tried to send a problem report related to the declined proof, but sending a problem report requires an active connection, which no longer existed.  This change prevents that error from occurring.

n/a

# Breaking change guide

n/a

# Related Issues

n/a

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/components/listItems/NotificationListItem.tsx
+++ b/packages/legacy/core/App/components/listItems/NotificationListItem.tsx
@@ -163,6 +163,11 @@ const NotificationListItem: React.FC<NotificationListItemProps> = ({
     try {
       const proofId = (notification as ProofExchangeRecord).id
       if (agent) {
+        const connection = await agent.connections.findById(proofId)
+        if (connection) {
+          await agent.proofs.sendProblemReport({ proofRecordId: proofId, description: t('ProofRequest.Declined') })
+        }
+
         await agent.proofs.declineRequest({ proofRecordId: proofId })
       }
     } catch (err: unknown) {
@@ -268,11 +273,11 @@ const NotificationListItem: React.FC<NotificationListItemProps> = ({
 
           // message.comment is the common fallback title for both v1 and v2 proof requests
           let body: string = message?.comment ?? ''
-          
+
           if (message instanceof V1RequestPresentationMessage) {
             body = message.indyProofRequest?.name ?? body
           }
-          
+
           if (message instanceof V2RequestPresentationMessage) {
             // workaround for getting proof request name in v2 proof request
             // https://github.com/openwallet-foundation/credo-ts/blob/5f08bc67e3d1cc0ab98e7cce7747fedd2bf71ec1/packages/core/src/modules/proofs/protocol/v2/messages/V2RequestPresentationMessage.ts#L78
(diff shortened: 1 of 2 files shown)


packages/legacy/core/App/components/listItems/NotificationListItem.tsx
```diff
@@ -163,6 +163,11 @@ const NotificationListItem: React.FC<NotificationListItemProps> = ({
     try {
       const proofId = (notification as ProofExchangeRecord).id
       if (agent) {
+        const connection = await agent.connections.findById(proofId)
```
bryce-mcmath: This method expects a `connectionId` no? Will a `proofId` work as expected? Are the two always the same?

https://github.com/openwallet-foundation/credo-ts/blob/9f404f2c67495df3c24d3e63040bcf2349fb4e6a/packages/core/src/modules/connections/ConnectionsApi.ts#L470

packages/legacy/core/App/screens/ProofRequest.tsx
```diff
@@ -561,7 +561,11 @@ const ProofRequest: React.FC<ProofRequestProps> = ({ navigation, proofId }) => {
   const handleDeclineTouched = useCallback(async () => {
     try {
       if (agent && proof) {
-        await agent.proofs.sendProblemReport({ proofRecordId: proof.id, description: t('ProofRequest.Declined') })
+        const connection = await agent.connections.findById(proof.id)
```
bryce-mcmath: "

packages/legacy/core/App/components/listItems/NotificationListItem.tsx
al-rosenthal: `ProofExchangeRecord` has a `connectionId`

jleach: Oof. Yes, not sure what I was thinking here.

al-rosenthal (approved):

bryce-mcmath:
If there ends up being any issue remaining for this we may just have to add an `.assertReady()` check on the connection as well.

packages/legacy/core/App/screens/ProofRequest.tsx
```diff
@@ -561,11 +561,17 @@ const ProofRequest: React.FC<ProofRequestProps> = ({ navigation, proofId }) => {
...
+
+        if (connection) {
+          await agent.proofs.sendProblemReport({ proofRecordId: proof.id, description: t('ProofRequest.Declined') })
+        }
+
         await agent.proofs.declineRequest({ proofRecordId: proof.id })
 
-        if (proof.connectionId && goalCode && goalCode.endsWith('verify.once')) {
-          agent.connections.deleteById(proof.connectionId)
+        if (connectionId && goalCode && goalCode.endsWith('verify.once')) {
```
bryce-mcmath, line 573: @jleach I think another check for whether the connection exists or not is needed here to prevent the error in bcgov/bc-wallet-mobile#2349

jleach: On like 565 above I fetch the connection and make sure it exists. Good enough?

bryce-mcmath: How would you feel about something like this for line 573:

```suggestion
        if (connection && goalCode?.endsWith('verify.once')) {
```

Not too aggressive of a change

PR #1439, opened 2025-02-12
feat: upgrade to RN 0.73.6

MosCD3:
# Summary of Changes

iOS part done so far
All changes re-reviewed and re-done by @MaSMas0  

New version supports experimental and chrome debugger (RN Debugger)
Added menu to start chrome debugger

Simulator Screenshot - iPhone 16 - 2025-02-18 at 18 00 27

Screenshot 2025-02-18 at 5 54 54 PM

Replace this text with embedded media for UI changes if they are included in this PR. If there are none, simply enter N/A

# Breaking change guide

# Related Issues

# Pull Request Checklist

- [ ] All commits contain a DCO `Signed-off-by` line (we use the [DCO GitHub app](https://github.com/apps/dco) to enforce this)

Pro Tip 🤓


bryce-mcmath:
@MosCD3 could we possibly chat about this PR at the Bifold call next week?

.github/workflows/main.yaml
```diff
@@ -66,10 +66,21 @@ jobs:
...
       - name: Install iOS dependencies
         # if: steps.pod-cache.outputs.cache-hit != 'true' || steps.npm-cache.outputs.cache-hit != 'true'
         working-directory: ./packages/legacy/app
         run: |
+          bundle install && \
+          cd ios && \
+          rm -rf Pods && \
+          rm -f Podfile.lock && \
+          pod install && \
+          cd .. && \
```
bryce-mcmath: Erasing the lockfile on the main GHA defeats the purpose of a lockfile. This change would stop our build from being deterministic . Can we make use of the same command (`ios:setup`) we would use locally? It's alright if we have to update that command a bit as part of this PR but we should still use it.

packages/legacy/app/Gemfile
```diff
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
 # You may use http://rbenv.org/ or https://rvm.io/ to install and use this version
 ruby ">= 2.6.10"
-gem 'cocoapods', '~> 1.12'
-gem 'activesupport', '~> 7.0.8'
+gem 'cocoapods', '>= 1.13', '< 1.15'
+gem 'activesupport', '>= 6.1.7.5', '< 7.1.0'
```
bryce-mcmath:
```suggestion
gem 'cocoapods', '~> 1.14.3'
gem 'activesupport', '~> 7.0.8'
```
This could work, no? More precise, less potential variation in builds? Would still pick up patch version bumps but no minor or major ones.

packages/legacy/core/App/contexts/auth.tsx
```diff
@@ -1,9 +1,22 @@
...
+  // console.log('Askar native module not loaded in debug mode')
+}
+
+//NOTE: This line is throwing an error in debugging due to native modules is it still needed? maybe the above is a workaround
+// to be able to use RNDebugger in debug mode
 // eslint-disable-next-line import/no-extraneous-dependencies
 import 'reflect-metadata'
 
-import { AskarWallet } from '@credo-ts/askar'
+import { AskarWallet } from '@credo-ts/askar' // is this enough instead of line number 2
```
bryce-mcmath: I don't think we should conditionally import. Is the `NOTE` for the conditional import above or the reflect-metadata line below?

I don't *think* reflect-metadata is still needed but we can just test out the app after removing it to find out if that's true. We should remove it from the entire project if possible

packages/legacy/core/__tests__/components/__snapshots__/VerifierCredentialCard.test.tsx.snap
```diff
@@ -203,7 +203,7 @@ exports[`VerifierCredentialCard Component Renders correctly 1`] = `
           removeClippedSubviews={false}
           renderItem={[Function]}
           scrollEnabled={false}
-          scrollEventThrottle={50}
+          scrollEventThrottle={0.0001}
```
bryce-mcmath, line 206: What caused this change?

bryce-mcmath (changes requested):
Few things, but awesome work

packages/legacy/core/App/contexts/auth.tsx
MosCD3: @hyperledger/aries-askar-react-native
is that needed Bryce, we already importing AzkarWallet from credo
This import is causing issues with real native chrome debugger, if its not needed anymore its better to remove the whole statement

packages/legacy/core/__tests__/components/__snapshots__/VerifierCredentialCard.test.tsx.snap
MosCD3: this was part of the snapshot update

packages/legacy/app/Gemfile
MosCD3: I believe it could

packages/legacy/core/App/contexts/auth.tsx
bryce-mcmath: Probably fine, we should just check that without that import we are still able to accept credentials and view / delete them after an app restart

timbl-ont:
Authentication with a PIN is broken - react-native-keychain needs to be updated to at least  8.1.2 for RN 7.3 support. There is an Android bug fixed in 8.1.3 that allowed Biometric authentication to be bypassed if someone had recently authenticated.

There are also some security enhancements in later releases that should be considered for a future PR.

https://github.com/oblador/react-native-keychain/releases

jleach:
Decided on the call today to update this to 0.73.6.

al-rosenthal:
Looks like the iOS build is failing. Any status on the android side of the update @MosCD3?

packages/legacy/core/__tests__/components/__snapshots__/VerifierCredentialCard.test.tsx.snap
MaSMas0: hello bryce, 
sorry for the late reply ,
so for the change in snapshot it is related to how react native is implementing virualized list 
so here is the version that you had for v0.72.17:
https://github.com/facebook/react-native/blob/v0.72.17/packages/virtualized-lists/Lists/VirtualizedList.js
to cut the story short check 
https://github.com/facebook/react-native/blob/6304cce015e4260b914aafd98d37ccdf5c3f2150/packages/virtualized-lists/Lists/VirtualizedList.js#L118C1-L121C2

which states that the default for scrollEventThrottle is 50 

and here is the implementation of VirtualizedList in v0.73.6:
https://github.com/facebook/react-native/blob/v0.73.6/packages/virtualized-lists/Lists/VirtualizedList.js
which is stating the default value to be 0.0001 check:
https://github.com/facebook/react-native/blob/0384caad3458f7658e68653a0ab62fd6cc3e4483/packages/virtualized-lists/Lists/VirtualizedList.js#L1078

and hence the change was done in the snapshot

.github/workflows/main.yaml
MosCD3: @MaSMas0

packages/legacy/core/App/contexts/auth.tsx
MosCD3: I tested the flow works fine without this import

packages/legacy/app/Gemfile
MaSMas0: @bryce-mcmath  resolved according to the versions you suggested

.github/workflows/main.yaml
MaSMas0: @bryce-mcmath  resolved using the ios:setup command and removed the line of removing the lockfile

packages/legacy/core/App/contexts/auth.tsx
bryce-mcmath: I'm good to get rid of it then, thanks for checking 👍

.github/workflows/main.yaml
```diff
@@ -18,6 +18,9 @@ on:
 jobs:
   build-ios:
     runs-on: macos-latest
+    env:
+      # Explicitly tell React Native where to find Node
+      REACT_NATIVE_NODE_BINARY: /Users/runner/hostedtoolcache/node/18.18.2/arm64/bin/node
```
bryce-mcmath: Is this still needed with the `$(which node)` below? Using `which` seems like the less brittle approach to me

packages/legacy/app/android/app/src/debug/AndroidManifest.xml
```diff
@@ -1,12 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools">
-    <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
```
bryce-mcmath, line 4: Just wondering, why was this removed?

packages/legacy/app/ios/Podfile
```diff
@@ -66,16 +66,11 @@ target 'AriesBifold' do
...
-  flags = get_default_flags()
 
   use_react_native!(
     :path => config[:reactNativePath],
-    # Hermes is now enabled by default. Disable by setting this flag to false.
-    :hermes_enabled => flags[:hermes_enabled],
-    :fabric_enabled => flags[:fabric_enabled],
     # Enables Flipper.
-    #
+    # shall we Enable Flipper?
```
bryce-mcmath: I don't think anyone uses Flipper on this project or that it was ever fully enabled. Safe to remove completely I think

packages/legacy/app/package.json
```diff
@@ -15,8 +15,8 @@
     "gradle:write-locks": "cd android && ./gradlew app:dependencies --write-locks",
     "prepare": "patch-package",
     "ios:setup": "bundle install --frozen && cd ios && bundle exec pod install",
-    "ios": "react-native run-ios",
-    "start": "react-native start",
+    "ios": "react-native run-ios --scheme AriesBifold",
+    "start": "react-native start --experimental-debugger",
```
bryce-mcmath, line 19: If the debugging code will be added in a future PR, should we remove this flag?

packages/legacy/core/__tests__/screens/UseBiometry.test.tsx
```diff
@@ -65,18 +65,25 @@ describe('UseBiometry Screen', () => {
...
-    await waitFor(() => {
-      timeTravel(1000)
-    })
+    // no need for the waitFor  it might cause infinte loop and cause the test to fail
+    // await waitFor(() => {
+    //   timeTravel(1000)
+    // })
+
+    // use this instead since anyway u r waiting the promise to resolve to set setBiometryAvailable in the useEffect
+    // this is cleaner approach than advanceTimersByTime
```
bryce-mcmath: Agree, can remove this comment and use the `findBy` approach

packages/legacy/core/__tests__/screens/UseBiometry.test.tsx
```diff
@@ -86,20 +93,25 @@ describe('UseBiometry Screen', () => {
...
 
-    await waitFor(() => {
-      timeTravel(1000)
-    })
+    // await waitFor(() => {
+    //   timeTravel(1000)
+    // })
+    // await act(async () => {
+    //   await new Promise((resolve) => setTimeout(resolve, 0))
+    // })
```
bryce-mcmath: Can remove this as well

packages/legacy/core/__tests__/screens/UseBiometry.test.tsx
```diff
@@ -86,20 +93,25 @@ describe('UseBiometry Screen', () => {
...
+
+    await tree.findByText('Biometry.NotEnabledText1')
 
     expect(tree).toMatchSnapshot()
   })
 
   test('Toggles use biometrics ok', async () => {
-    authContext.isBiometricsActive = jest.fn().mockResolvedValueOnce(true)
+    // authContext.isBiometricsActive = jest.fn().mockResolvedValueOnce(true)
```
bryce-mcmath: Can remove

packages/legacy/core/__tests__/screens/W3cProofRequest.test.tsx
```diff
@@ -128,12 +128,12 @@ describe('displays a proof request screen', () => {
         </BasicAppContext>
       )
 
-      await waitFor(() => {
-        timeTravel(1000)
-      })
+      // await waitFor(() => {
+      //   timeTravel(1000)
+      // })
```
bryce-mcmath: Can remove

MaSMas0:
# PR Summary: React Native Upgrade from v0.72.17 to v0.73.6

Based on the file structure you've shared, I can see this PR implements a significant upgrade of the Bifold Wallet from React Native v0.72.17 to v0.73.6. Here's a summary of the changes:

## Core Changes

- **JavaScript/TypeScript Migration**:

  - Upgraded React Native core dependencies in `package.json`
  - Updated various app context and screen files
  - Fixed `OnboardingStyleSheet` typing to use `ViewStyle` instead of `Record<string, any>` for type safety

- **Android Updates**:

  - Migrated Java files to Kotlin (`MainActivity.java` → `MainActivity.kt`, `MainApplication.java` → `MainApplication.kt`)
  - Updated `build.gradle` and `gradle.lockfile` configurations
  - Modified `AndroidManifest.xml` for compatibility with newer RN version
  - Updated Gradle version to be compatible with RN 0.73.6

- **iOS Updates**:
  - Updated `Podfile` and `Podfile.lock` with new dependency versions
  - Modified XCode project settings for compatibility
  - Changed `AppDelegate` from Objective-C to Objective-C++ for compatibility
  - Updated the Gemfile to be compatible with the new RN version
  - Fixed `yarn ios` script to explicitly choose the valid iOS schema

## Build & Configuration Updates

- Updated GitHub workflow files (`main.yaml`) for iOS and Android builds
- Added `.nvmrc` to standardize Node.js version across development environments
- Modified various babel configurations across packages
- Added react-native-reanimated to babel.config.js as recommended by official documentation
- Added the experimental debugger for the `yarn start` script (recommended for Hermes)
- Ensured Hermes is properly configured in Podfile

## Testing Infrastructure

- Updated test snapshots to accommodate UI changes from the new React Native version
- Revised Jest configuration to work with the upgraded environment
- Refactored unit tests to address issues with time-based tests:
  - Fixed error-prone code using waitFor with time travel
  - Implemented FakeTimers for long-waited timeouts
  - Improved handling of components affected by upgrade (RNCSafeAreaView, RNSVGRect, VirtualizedList)

## Developer Experience

- Added `.nvmrc` for consistent Node.js version management
- For zsh users, this script can be added to automatically use the correct Node version:

```bash
# Add this to your .zshrc
autoload -U add-zsh-hook
load-nvmrc() {
  local node_version="$(nvm version)"
  local nvmrc_path="$(nvm_find_nvmrc)"

  if [ -n "$nvmrc_path" ]; then
    local nvmrc_node_version=$(nvm version "$(cat "${nvmrc_path}")")

    if [ "$nvmrc_node_version" = "N/A" ]; then
      nvm install
    elif [ "$nvmrc_node_version" != "$node_version" ]; then
      nvm use
    fi
  elif [ "$node_version" != "$(nvm version default)" ]; then
    echo "Reverting to nvm default version"
    nvm use default
  fi
}
add-zsh-hook chpwd load-nvmrc
load-nvmrc
```


For more information and support for other platforms, refer to https://github.com/nvm-sh/nvm#zsh

This upgrade provides the benefits of React Native 0.73.6, including improved performance, bug fixes, and support for newer platform features, while maintaining the core functionality of the Bifold Wallet.

bryce-mcmath (changes requested):
Some small comments. If it's easier I can make these changes if you invite me to collab on this branch

packages/legacy/app/android/app/src/debug/AndroidManifest.xml
MaSMas0: @bryce-mcmath  you can read this https://github.com/facebook/react-native/issues/5886#issuecomment-464495388 about 
Removing unused permissions added at build time and this common unused permission you usually use it in debugging mode

packages/legacy/app/package.json
MaSMas0: @bryce-mcmath  i dont think we should as it is anyway working what i removed is the workaround for the remote debugging but the experimental is the default for hermes so no i think we shouldnt but if you want we can add a new script like start:debug

bryce-mcmath: Should be fine then 👍

packages/legacy/app/android/app/src/debug/AndroidManifest.xml
bryce-mcmath: Ah nice, good catch 👍

packages/legacy/core/__tests__/screens/W3cProofRequest.test.tsx
MaSMas0: done!

packages/legacy/core/__tests__/screens/UseBiometry.test.tsx
MaSMas0: done!

packages/legacy/core/__tests__/screens/UseBiometry.test.tsx
MaSMas0: done!

packages/legacy/core/__tests__/screens/UseBiometry.test.tsx
MaSMas0: done!

packages/legacy/app/ios/Podfile
MaSMas0: that is just a comment and it is removed

.github/workflows/main.yaml
MaSMas0: i think `$(which node)` is sufficient and yes it is less brittle but i was trying to find out why the GHA was failing so i was ruling this out

packages/legacy/app/ios/ariesbifold/AppDelegate.mm
bryce-mcmath, line 1: What was the reason for this change? Going from `.m` to `.mm` I mean

MaSMas0: @bryce-mcmath  i mentioned this in the PR Summary   https://github.com/openwallet-foundation/bifold-wallet/pull/1439#issuecomment-2722719952

Changed AppDelegate from Objective-C to Objective-C++ for compatibility

please check the summary to know more

.github/workflows/main.yaml
```diff
@@ -66,17 +69,30 @@ jobs:
         run: |
           yarn lerna run build
 
+      - name: Create .xcode.env.local file
+        working-directory: ./packages/legacy/app/ios
+        run: |
+          echo "export NODE_BINARY=$(which node)" > .xcode.env.local
+          cat .xcode.env.local
```
bryce-mcmath, line 73: Is this needed or was it just added for troubleshooting?

.github/workflows/main.yaml
```diff
@@ -66,17 +69,30 @@ jobs:
...
+
       - name: Install iOS dependencies
         # if: steps.pod-cache.outputs.cache-hit != 'true' || steps.npm-cache.outputs.cache-hit != 'true'
         working-directory: ./packages/legacy/app
+        env:
+          REACT_NATIVE_NODE_BINARY: $(which node)
+          NODE_BINARY: $(which node)
```
bryce-mcmath: Are these needed here as well as on line 21? It should still work if they were both set up top only, no?

.github/workflows/main.yaml
```diff
@@ -66,17 +69,30 @@ jobs:
...
           yarn run ios:setup && \
           git status && \
           git diff ios/Podfile.lock
 
       - name: Run release build
         # if: steps.pod-cache.outputs.cache-hit != 'true' || steps.npm-cache.outputs.cache-hit != 'true'
         working-directory: ./packages/legacy/app/ios
+        env:
+          REACT_NATIVE_NODE_BINARY: $(which node)
+          NODE_BINARY: $(which node)
```
bryce-mcmath: Same comment as above

bryce-mcmath (commented):
Couple more small things related to the GHA that the summary comment didn't clear up for me, sorry to be a bother

bryce-mcmath:
@MaSMas0 @MosCD3 the only thing (other than the GHA stuff) that I think should be changed before this PR is good to go, is `react-native-reanimated` should be pegged to version `3.6.3` rather than `^3.6.0` because I found that other packages that use it will sometimes resolve it to be a higher minor version that RN 0.73.6 doesn't work with (eg. https://github.com/software-mansion/react-native-reanimated/blob/bd4752b55935d3057d21d8141f98c2e6e2dae9c7/packages/react-native-reanimated/scripts/reanimated_utils.rb#L56-L57)

Might even be worth adding to the resolutions block as well

MaSMas0:
@bryce-mcmath for the react-native-reanimated i dont see the issue and how it is related to carret as all minor versions of react-native-reanimated is supported by RN 0.73.6 as mentioned here in this link : https://docs.swmansion.com/react-native-reanimated/docs/guides/compatibility

image

as for this https://github.com/software-mansion/react-native-reanimated/blob/bd4752b55935d3057d21d8141f98c2e6e2dae9c7/packages/react-native-reanimated/scripts/reanimated_utils.rb#L56-L57)
this is related to this PR https://github.com/software-mansion/react-native-reanimated/pull/6995
which states that it is related to Reanimated v4 & v3.17 which is way higher than v3.6.0

image

P.S: react-native-reanimated isnt really used in the project at all maybe it is planned to be but right now it exist as additional dependency but has no real usage in the app so maybe it should be removed till you decide to use it instead of react native built in animation

.github/workflows/main.yaml
MosCD3: @MaSMas0

.github/workflows/main.yaml
MosCD3: @MaSMas0

.github/workflows/main.yaml
MosCD3: @MaSMas0

MosCD3:
@bryce-mcmath I think we should remove the re-animated package if its not directly used

bryce-mcmath:
@MaSMas0 

>it is related to Reanimated v4 & v3.17 which is way higher than v3.6.0

In semantic versioning, a package with version x.y.z means x is the major version, y is the minor, and z is the patch. A tilde (~) will allow packages to upgrade the patch version automatically, and a caret (^) will allow packages to upgrade the minor and patch versions automatically. In this example, it means `^3.6.0` can get bumped to `3.17.0`. Which is not supported by RN versions lower than 0.75. Adding it in the resolutions block is also helpful as it enforces that version even when it is used as a transient dependency

> P.S: react-native-reanimated isnt really used in the project at all maybe it is planned to be but right now it exist as additional dependency but has no real usage in the app so maybe it should be removed till you decide to use it instead of react native built in animation

I'm fairly sure even if it is not used directly it is required to support other packages we use

MaSMas0:
@bryce-mcmath 
> In semantic versioning, a package with version x.y.z means x is the major version, y is the minor, and z is the patch. A tilde (~) will allow packages to upgrade the patch version automatically, and a caret (^) will allow packages to upgrade the minor and patch versions automatically. In this example, it means `^3.6.0` can get bumped to `3.17.0`. Which is not supported by RN versions lower than 0.75. Adding it in the resolutions block is also helpful as it enforces that version even when it is used as a transient dependency
> 
you are right i confused the patch with minor :D and tilde with caret ok no problem

.github/workflows/main.yaml
MaSMas0: removed

.github/workflows/main.yaml
MaSMas0: done and removed

.github/workflows/main.yaml
MaSMas0: yes it is needed and cant be removed

MaSMas0:
@bryce-mcmath 
> I'm fairly sure even if it is not used directly it is required to support other packages we use

i am not sure that this is the case  can you please check as from my side i dont see which package is depending on react-native-reanimated 

more clarification would be of a help as this is too big of package to be existing without a usage

bryce-mcmath:
@MaSMas0 
> i am not sure that this is the case can you please check as from my side i dont see which package is depending on react-native-reanimated

It's a dependency and peer dependency of `react-native-vision-camera` which is core to the app, so it's required and would be installed anyway even if we didn't list it explicitly, only the version would be whatever that `react-native-vision-camera` package prefers (3.17.x) which would break the app due to the reason mentioned above

MaSMas0:
@bryce-mcmath 
> It's a dependency and peer dependency of `react-native-vision-camera` which is core to the app, so it's required and would be installed anyway even if we didn't list it explicitly, only the version would be whatever that `react-native-vision-camera` package prefers (3.17.x) which would break the app due to the reason mentioned above
>
i think that this is not the case as you can see the peerDependenciesMeta showing that it is optional that the package wont break the app  for example you can see @shopify/react-native-skia and you dont have it in project and yet the app works coz it is optional so i still think that you dont need it also if u check the docs of react-native-vision-camera it doesnt state that you need  react-native-reanimated as peer dependencies in order for the package to work correctly so i suggest to remove it from the project as long as it is not needed

`"react-native-vision-camera@npm:4.3.1":
  version: 4.3.1
  resolution: "react-native-vision-camera@npm:4.3.1"
  peerDependencies:
    "@shopify/react-native-skia": "*"
    react: "*"
    react-native: "*"
    react-native-reanimated: "*"
    react-native-worklets-core: "*"
  peerDependenciesMeta:
    "@shopify/react-native-skia":
      optional: true
    react-native-reanimated:
      optional: true
    react-native-worklets-core:
      optional: true
  checksum: e97ad8503cfd848cc794f95252b5f6385b28cf16c3606c3852654c9ee623288d7e820e3dc95f75e332d31b9a8d01090de281f3bf124d97da8dc786253b1c1283
  languageName: node
  linkType: hard
`

bryce-mcmath:
@MaSMas0 
> `"react-native-vision-camera@npm:4.3.1": version: 4.3.1 resolution: "react-native-vision-camera@npm:4.3.1" peerDependencies: "@shopify/react-native-skia": "*" react: "*" react-native: "*" react-native-reanimated: "*" react-native-worklets-core: "*" peerDependenciesMeta: "@shopify/react-native-skia": optional: true react-native-reanimated: optional: true

ohhh nice catch! ok let's remove it but if, after we remove it, it still gets installed as a transient dependency by another one of our deps, can we add an entry to the root `package.json` like this?

```
"resolutions": {
  ...
  "react-native-reanimated": "3.6.3"
  ...
}
```

MaSMas0:
@bryce-mcmath 
> ohhh nice catch! ok let's remove it but if, after we remove it, it still gets installed as a transient dependency by another one of our deps, can we add an entry to the root `package.json` like this?
> 
> ```
> "resolutions": {
>   ...
>   "react-native-reanimated": "3.6.3"
>   ...
> }
> ```

Yes, that's exactly the right approach. If react-native-reanimated is being installed as a transitive dependency from another package. we can  surely use the resolutions field in root package.json to control its version

but lets make this when we have  that case i guess right now it is ok to move on with PR as we removed react-native-reanimated

MaSMas0:
@bryce-mcmath 
> ohhh nice catch! ok let's remove it but if, after we remove it, it still gets installed as a transient dependency by another one of our deps, can we add an entry to the root `package.json` like this?
> 
> ```
> "resolutions": {
>   ...
>   "react-native-reanimated": "3.6.3"
>   ...
> }
> ```

Yes, that's exactly the right approach. If react-native-reanimated is being installed as a transitive dependency from another package. we can  surely use the resolutions field in root package.json to control its version

but lets make this when we have a that case i guess right now i guess it is ok to move on with PR as we removed react-native-reanimated

bryce-mcmath (approved):
Will approve again after conflicts are fixed but really great work! Thanks for your patience with all the staggered feedback

MosCD3:
Nice work all of you

PR #1440, opened 2025-02-13
feat: add scan overlay, refactor

bryce-mcmath:
# Summary of Changes

This PR adds an overlay to the QR code scanner inspired by PR #1420 and refactors two duplicate components we had into one (`QRScanner.tsx` and `NewQRView.tsx`). A few additional tests were also added

As a little side quest I added a readme file for patch notes

### Before, with and without the connection inviter capability enabled from developer mode:
QRScanner
NewQRView

### After, with and without the connection inviter capability enabled from developer mode:
QRScanner_New
QRScanner_New_ConnectionInviterEnabled

# Breaking change guide

N/A

# Related Issues

#1420
bcgov/bc-wallet-mobile#2377

# Pull Request Checklist

Pro Tip 🤓


bryce-mcmath:
@timbl-ont @jian4on FYI

timbl-ont:
Looks good. One question is the size of the scan area. It is bigger than the pervious screens. Was that intentional?

bryce-mcmath:
@timbl-ont 
> Looks good. One question is the size of the scan area. It is bigger than the pervious screens. Was that intentional?

Yes, on some phones (Android especially) denser QR codes don't always scan until you are closer than the previous scan area implies you need to be. But I'll defer to the UX folks if they would like it to be changed

knguyenBC:
I think the increased scan target is fine. We have had issues with dense QR codes and try to reduce them where we can. Ideally QR codes are not so dense but I suppose this is minor improvement. I assume that QR codes aren't required to fill the space and that the qr code will scan regardless.

bryce-mcmath:
@knguyenBC
> I assume that QR codes aren't required to fill the space and that the qr code will scan regardless.

That is correct, the actual scan mechanism works the same regardless of the overlay

packages/legacy/core/App/components/misc/QRScanner.tsx
```diff
@@ -65,105 +121,267 @@ const QRScanner: React.FC<Props> = ({ handleCodeScan, error, enableCameraOnError
...
-            >
-              <Icon name="help-circle" size={24} style={{ color: 'white' }} />
-            </Pressable>
-          )}
-
-          <View style={{ width: 10, marginLeft: 'auto' }} />
-          <QRScannerTorch active={torchActive} onPress={() => setTorchActive(!torchActive)} />
-        </View>
-      </View>
+        {showTabs ? (
```
al-rosenthal, line 367: could update to `showTabs ?? (...)`

packages/legacy/core/App/components/misc/QRScanner.tsx
```diff
@@ -65,105 +121,267 @@ const QRScanner: React.FC<Props> = ({ handleCodeScan, error, enableCameraOnError
...
+            <ScanTab
+              title={t('Scan.MyQRCode')}
+              iconName={'qr-code'}
+              onPress={() => setFirstTabActive(false)}
+              active={!firstTabActive}
+            />
+          </View>
+        ) : null}
+      </SafeAreaView>
+      {showTabs ? <SafeAreaView edges={['bottom']} style={styles.bottomSafeArea}/> : null}
```
al-rosenthal, line 384: could update to `showTabs ?? (...)`

packages/legacy/core/App/components/misc/QRScanner.tsx
bryce-mcmath: The docs seem to recommend either `condition && A` or `condition ? A : B`
https://react.dev/learn/conditional-rendering

I usually use `?` in case the condition (eg. `showTabs`) is ever `0` by some accident - if we use `&&` (and I believe `??` as well) it will actually try to render the text `0` in that case. If we use `condition ? A : null` it's a little sturdier

packages/legacy/core/App/components/misc/QRScanner.tsx
bryce-mcmath: "

al-rosenthal (approved):

PR #1441, opened 2025-02-13
fix: disable header elements and bottom nav for TalkBack during tours

ClaudeArs:
# Summary of Changes

Disabled header elements and bottomNav for talkBack in ModalTour

N/A

# Breaking change guide
N/A

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/components/buttons/HeaderHome.tsx
+++ b/packages/legacy/core/App/components/buttons/HeaderHome.tsx
@@ -1,20 +1,29 @@
 import { useNavigation } from '@react-navigation/native'
-import React from 'react'
+import React, { useEffect, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 
 import { Screens, TabStacks } from '../../types/navigators'
 import { testIdWithKey } from '../../utils/testable'
 
 import IconButton, { ButtonLocation } from './IconButton'
+import { useTour } from '../../contexts/tour/tour-context'
+import { ImportantForAccessibility } from '../../types/accessibility'
 
 const HeaderRightHome: React.FC = () => {
   const { t } = useTranslation()
   const navigation = useNavigation()
+  const { currentStep } = useTour()
+  const [hideElements, setHideElements] = useState<ImportantForAccessibility>('auto')
+
+  useEffect(() => {
+    setHideElements(currentStep === undefined ? 'auto' : 'no-hide-descendants')
+  }, [currentStep])
 
   return (
     <IconButton
       buttonLocation={ButtonLocation.Right}
       accessibilityLabel={t('Global.Home')}
+      importantForAccessibility={hideElements}
       testID={testIdWithKey('HomeButton')}
       onPress={() => navigation.getParent()?.navigate(TabStacks.HomeStack, { screen: Screens.Home })}
       icon="home"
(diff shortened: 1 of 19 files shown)


packages/legacy/core/App/components/misc/NewQRView.tsx
bryce-mcmath, line 1: This file is going to be deleted soon (#1440)

packages/legacy/core/App/components/buttons/SettingsMenu.tsx
```diff
@@ -1,21 +1,31 @@
...
 const SettingsMenu: React.FC = () => {
   const navigation = useNavigation<StackNavigationProp<RootStackParams>>()
   const { t } = useTranslation()
 
+  const { currentStep } = useTour()
+  const [hideElements, setHideElements] = useState<ImportantForAccessibility>('auto')
+
+  useEffect(() => {
+    setHideElements(currentStep === undefined ? 'auto' : 'no-hide-descendants')
+  }, [currentStep])
```
bryce-mcmath, line 22: this might be a better place for `useMemo` rather than `useEffect`

packages/legacy/core/App/components/buttons/HeaderHome.tsx
```diff
@@ -1,20 +1,29 @@
...
 
 const HeaderRightHome: React.FC = () => {
   const { t } = useTranslation()
   const navigation = useNavigation()
+  const { currentStep } = useTour()
+  const [hideElements, setHideElements] = useState<ImportantForAccessibility>('auto')
+
+  useEffect(() => {
+    setHideElements(currentStep === undefined ? 'auto' : 'no-hide-descendants')
+  }, [currentStep])
```
bryce-mcmath, line 20: same thought here, this might be a better place for `useMemo` rather than `useEffect`. I wonder though, would it be simpler to set this `importantForAccessibility` property on the parent (header itself)? Then we wouldn't need to add it to all these children

packages/legacy/core/App/navigators/TabStack.tsx
```diff
@@ -102,14 +106,21 @@ const TabStack: React.FC = () => {
     [agent, enableImplicitInvitations, enableReuseConnections, logger, navigation, t, dispatch]
   )
 
+  useEffect(() => {
+    setHideElements(currentStep === undefined ? 'auto' : 'no-hide-descendants')
+  }, [currentStep])
```
bryce-mcmath, line 111: " (`useMemo`)

bryce-mcmath (changes requested):
Probably better (and less code) for us to add the property to the header component itself that contains all these others, besides `TabStack`. Another thing, do we not also need to set `accessibilityElementsHidden` for iOS?

packages/legacy/core/App/navigators/TabStack.tsx
ClaudeArs: Thanks for the comments. For iOS the accessibilityViewIsModal property in TourOverlay does the job but is not supported by talkBack that's why we have to disable the underlying elements.

Indeed, doing the code only once in the header component would be more judicious. I'll look into that

thanks

PR #1448, opened 2025-02-20
fix: prevent unnecessary network toasts

bryce-mcmath:
# Summary of Changes

This PR fixes an issue where the "No internet connection" toast would appear on cold start when the first network check still hadn't completed

In addition, I refactored the relevant code and added tests

N/A

# Breaking change guide

`silentAssertConnectedNetwork` can now return `boolean` or `null` - it will return null until the first check is complete

`assertInternetReachable` is now a synchronous function that returns `boolean` or `null` - there were no promises in the existing function so it was unnecessary for it to be `async` and the return was unsafely typecasted as a `boolean` when it can indeed be `null`

# Related Issues

bcgov/bc-wallet-mobile#2382

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/components/network/NetInfo.tsx
```diff
@@ -12,63 +11,57 @@ const NetInfo: React.FC = () => {
...
-          } else {
-            // Network is available but cannot access nediator, display toast
-            _showNetworkWarning()
-          }
-        })
-        return
-      } else if (!hasShown) {
-        _showNetworkWarning()
+    // Only check general internet connection if mediator check is disabled
+    if (disableMediatorCheck) {
```
jleach: I though a change was implemented to make the check generic because some teams don't use a mediator?

packages/legacy/core/App/components/network/NetInfo.tsx
```diff
@@ -12,63 +11,57 @@ const NetInfo: React.FC = () => {
...
+      const internetReachable = assertInternetReachable()
```
jleach: Wondering if we need a mediator check at all with this? Can we just check whatever is set?

jleach (changes requested):

jleach:
@bryce-mcmath Clarifying my comments. I though there was a change introduced to allow the "internet check" to check the default URL which may be Google or CloudFlare or whatever or, allow that URL to be overrides. So, for example, if we want to check the mediator we just override the default URL. If that's true then "internetIsReachable" or whatever the test is becomes the only test we need. No mediator checking required.

packages/legacy/core/App/components/network/NetInfo.tsx
bryce-mcmath: That hadn't been done yet, but I've added that to this PR now

jleach (approved):

PR #1453, opened 2025-02-21
fix: refactored wallet auth 

al-rosenthal:
# Summary of Changes

Changes made to help mitigate and provide better feedback for a strange issue with the application hanging during initialization. 

- moved `walletSecret` up as a param for `initialize-agent.ts` functions
- added thrown errors for missing `walletSecret` properties
- cleaned up various logical functions

n/a

# Breaking change guide

# Related Issues

[BC-Wallet: 2374](https://github.com/bcgov/bc-wallet-mobile/issues/2374)

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/services/keychain.ts
```diff
@@ -142,22 +143,30 @@ export const loadWalletKey = async (title?: string, description?: string): Promi
...
   let key: WalletKey | undefined
-  let err = ''
+  let secret: WalletSecret | undefined
   try {
     salt = await loadWalletSalt()
     key = await loadWalletKey(title, description)
   } catch (e: any) {
-    err = e?.message ?? e
+    // should this throw instead?
+    DeviceEventEmitter.emit(EventTypes.BIOMETRY_ERROR, e?.message ?? e)
```
al-rosenthal: The original code captured this error and added it to the return. It emitted an error but continued on with an undefined wallet secret, I'm wondering if this should throw an error instead emitting an error

packages/legacy/core/App/services/keychain.ts
```diff
@@ -63,21 +63,21 @@ export const storeWalletKey = async (secret: WalletKey, useBiometrics = false):
   const secretAsString = JSON.stringify(secret)
   await wipeWalletKey(useBiometrics)
   const result = await Keychain.setGenericPassword(keyFauxUserName, secretAsString, opts)
-  return typeof result === 'boolean' ? false : true
+  return Boolean(result)
```
al-rosenthal, line 67: Boolean() converts the param into a truthy result, if the null is present it is false, if the object is present it returns true.

packages/legacy/core/App/contexts/auth.tsx
```diff
@@ -52,83 +50,79 @@ export const AuthProvider: React.FC<React.PropsWithChildren> = ({ children }) =>
...
+        if (!store.migration.didMigrateToAskar) {
+          await migrateToAskar(secret.id, hash)
+          dispatch({
+            type: DispatchAction.DID_MIGRATE_TO_ASKAR,
+          })
+        }
+
+        // NOTE: a custom wallet is used to check if the wallet key is correct. This is different from the wallet used in the rest of the app.
+        // We create an AskarWallet instance and open the wallet with the given secret.
+        const askarWallet = new AskarWallet(
```
jleach, line 107: I think this is an instance of the existing wallet. If so, it might be good to clarify the docs that it's not creating a new wallet, but just testing with an instance of the existing one.

packages/legacy/core/App/contexts/auth.tsx
```diff
@@ -52,83 +50,79 @@ export const AuthProvider: React.FC<React.PropsWithChildren> = ({ children }) =>
...
-      if (!store.migration.didMigrateToAskar) {
-        await migrateToAskar(secret.id, hash)
-        dispatch({
-          type: DispatchAction.DID_MIGRATE_TO_ASKAR,
-        })
+      // set did authenticate to true if we can get wallet credentials
+      dispatch({
+        type: DispatchAction.DID_AUTHENTICATE,
+      })
+      if (useBiometry) {
```
jleach: A few spaces around here might help with readability.

packages/legacy/core/App/contexts/auth.tsx
```diff
@@ -52,83 +50,79 @@ export const AuthProvider: React.FC<React.PropsWithChildren> = ({ children }) =>
...
-    }
-    return true
-  }, [dispatch, getWalletCredentials])
-
-  const checkPIN = useCallback(async (PIN: string): Promise<boolean> => {
-    try {
-      const secret = await loadWalletSalt()
-
-      if (!secret || !secret.salt) {
+  const commitPIN = useCallback(
```
jleach: Opportunity to clarify the name? commitWalletToKeychain or something?

packages/legacy/core/App/contexts/auth.tsx
```diff
@@ -138,29 +132,32 @@ export const AuthProvider: React.FC<React.PropsWithChildren> = ({ children }) =>
...
+          return false
+        }
+        const oldHash = await hashPIN(oldPin, secret.salt)
+        const newSecret = await secretForPIN(newPin)
+        const newHash = await hashPIN(newPin, newSecret.salt)
+        if (!newSecret.key) {
+          return false
+        }
+
+        await agent?.wallet.close()
```
jleach: I might just check agent at the top and return false if its undefined. No point id doing anything if its undefined.

packages/legacy/core/App/services/keychain.ts
jleach: Good to know. I think we have a few places in code where we use `!!`.

packages/legacy/core/App/services/keychain.ts
jleach: That event will trigger a global error catcher but if the wallet can't be unlocked the entire app is pretty useless. My though is no point in carrying on with an invalid secret.

jleach (changes requested):

packages/legacy/core/App/contexts/auth.tsx
bryce-mcmath: Another thing, shouldn't we be reopening the wallet after this line? I know this is what we had already been doing but I don't think we actually test this feature much so maybe it's broken. Tried getting a credential after changing your pin (without restarting)?

packages/legacy/core/App/services/keychain.ts
```diff
@@ -142,22 +143,30 @@ export const loadWalletKey = async (title?: string, description?: string): Promi
...
     salt = await loadWalletSalt()
     key = await loadWalletKey(title, description)
   } catch (e: any) {
-    err = e?.message ?? e
+    // should this throw instead?
+    DeviceEventEmitter.emit(EventTypes.BIOMETRY_ERROR, e?.message ?? e)
   }
 
-  return { secret: { ...salt, ...key } as WalletSecret, err }
+  if (salt && key) {
```
bryce-mcmath: Should we do some kind of validation that the `salt` has all the required properties?

packages/legacy/core/App/contexts/auth.tsx
```diff
@@ -52,83 +50,79 @@ export const AuthProvider: React.FC<React.PropsWithChildren> = ({ children }) =>
   }, [])
 
   const getWalletCredentials = useCallback(async (): Promise<WalletSecret | undefined> => {
-    if (walletSecret && walletSecret.key) {
+    if (walletSecret) {
```
bryce-mcmath, line 53: How are we sure it will have all the required properties at this point?

bryce-mcmath (commented):
Some questions. Might be a good opportunity to give these functions better names while we're at it

packages/legacy/core/App/contexts/auth.tsx
al-rosenthal: good call, I'll test that out. I agree though, no point in the agent trying anything if it isn't there

packages/legacy/core/App/contexts/auth.tsx
al-rosenthal: does the new comment make more sense to you?

packages/legacy/core/App/contexts/auth.tsx
al-rosenthal: WalletSecret is more strongly typed and anywhere the wallet value is being set will now throw or return the expected values

al-rosenthal: if you still have concerns I'm happy to add more explicit checks, I just wanted to keep things a little cleaner

bryce-mcmath: I love it

packages/legacy/core/App/utils/crypto.ts
```diff
@@ -1,8 +1,12 @@
...
-  const result = await argon2(PIN, salt, {})
-  const { rawHash } = result
+  try {
+    const result = await argon2(PIN, salt, {})
+    const { rawHash } = result
 
-  return rawHash
+    return rawHash
+  } catch (error) {
+    throw new Error('Error generating hash for PIN' + String(error))
```
bryce-mcmath: Nit for error readability: 
```suggestion
    throw new Error('Error generating hash for PIN: ' + String(error))
```

jleach: I think this is the old skool format for string concatenation. New way is with backticks and `${error.message}` or whatever.

packages/legacy/core/App/services/keychain.ts
```diff
@@ -142,22 +154,30 @@ export const loadWalletKey = async (title?: string, description?: string): Promi
 export const loadWalletSecret = async (
   title?: string,
   description?: string
-): Promise<{ secret: WalletSecret | undefined; err: string }> => {
+): Promise<WalletSecret | undefined> => {
   let salt: WalletSalt | undefined
   let key: WalletKey | undefined
-  let err = ''
+  let secret: WalletSecret | undefined = undefined
```
jleach, line 161: I wonder if having WalletSalt, WalletKey, and WalletSecret are over complicating? Perhaps not for this PR but if the opportunity presents that might be worth considering. To me WalletSalt and WalletKey are just `Partial<Type>` of WalletSecret like:

```
type User = {
  name: string;
  age: number;
  email: string;
}
```

```
const partialUser: Partial<User> = {
  name: 'Bob'
}; // Valid - `age` and `email` are optional
```

This may have been the case at some point which is why `key?` was optional. I noticed in this PR you changed it to required.

jleach (dismissed):
Comments at this point really.

packages/legacy/core/App/services/keychain.ts
al-rosenthal: good point, I've added a todo to try and capture this idea @jleach

bryce-mcmath (approved):

PR #1464, opened 2025-03-10
refactor: onboarding mechanics and navigation

jleach:
# Summary of Changes

### Improving Onboarding Flexibility  

Currently, adding or reorganizing onboarding screens is cumbersome due to hardcoded navigation logic. This PR simplifies the process by relying more on application state and less on manual navigation rules.  

### Key Changes:  
- Onboarding screens are now managed as an array, making it easier to modify the order of screens.  
- The `OnboardingStack` automatically displays the first incomplete screen.  
- Once all screens are completed, an event is emitted to signal that onboarding is finished, prompting the root stack to transition to `MainStack`.  
- The only required screens in this process are `CreatePIN` and/or `EnterPIN`.  

Replace this text with embedded media for UI changes if they are included in this PR. If there are none, simply enter N/A

# Breaking change guide

# Related Issues

# Pull Request Checklist

Pro Tip 🤓


packages/legacy/core/App/navigators/OnboardingStack.tsx
```diff
@@ -15,9 +28,11 @@ import { createCarouselStyle } from '../screens/OnboardingPages'
 import PINCreate from '../screens/PINCreate'
 import PINEnter from '../screens/PINEnter'
 import PushNotification from '../screens/PushNotification'
-import { AuthenticateStackParams, Screens } from '../types/navigators'
+// import UpdateAvailable from '../screens/UpdateAvailable'
```
fc-santos: @jleach from what I can see the BC team is working on the UpdateAvailable implementation. From what I heard, maybe @cvarjao can confirm this, but the bc wallet wants to make all new updates required. Would there be any flexibility in the bifold (with an injectable TOKEN) to decide if an update is mandatory/required or not ? In Quebec, we haven't decided yet if we want to make all updates required. And what happens if the app doesn't exist in the store yet ?

packages/legacy/core/App/container-impl.ts
```diff
@@ -52,6 +52,7 @@ import ContactCredentialListItem from './components/listItems/ContactCredentialL
 import { InlineErrorPosition } from './types/error'
 import { DefaultScreenLayoutOptions } from './navigators/defaultLayoutOptions'
 import ConnectionAlert from './components/misc/ConnectionAlert'
+import { DeviceEventEmitter } from 'react-native'
```
fc-santos: It looks like this import and the one added above are not being used in this file.

packages/legacy/core/App/navigators/OnboardingStack.tsx
jleach: We've noticed that some users continue to contact support about bugs that have already been fixed because they haven’t updated their app yet. To help guide them in the right direction, we’re introducing a reminder screen in React Native.

Initially, I was thinking of making this screen optional, but it makes sense to require updates. For now, the app will gently remind users to update when they open it. 😊

fc-santos: Is there any work on progress for blocking the user from using the app, if let's say, there is a required update the user needs for the app to continue working (e.g: a change in the mediator url). And would the implementation of that logic be more to each wallet to implement or could it be done in the bifold?

fc-santos: Also, would this screen only appear in the first install of the app ?

jleach: @fc-santos No to work in progress to block users. Also no to would it appear on the first install. The idea is it would check every time the app starts, and if an update is available, show the screen to the user. It kindly nag them.

packages/legacy/core/App/navigators/OnboardingScreens.ts
fc-santos, line 1: Shouldn't all options be configured in the ScreenOptionsDictionary instead of here, so other wallets can decide what they want?

Is there a reason why some elements of the ScreenOptions[] have options as a type of object and other elements the options is a function that returns an object?

packages/legacy/core/App/navigators/OnboardingScreens.ts
```diff
@@ -0,0 +1,140 @@
...
+  ScreenOptionsDictionary: ScreenOptionsType,
+  components: ScreenComponents
+): ScreenOptions[] => [
+  {
+    name: Screens.Splash,
+    component: components.Splash,
+    options: {
+      ...ScreenOptionsDictionary[Screens.Splash],
+      ...TransitionPresets.ModalFadeTransition,
+      title: t('Screens.Preface'),
```
fc-santos: Shouldn't this be Screens.Splash ?

jleach: Fixed.

packages/legacy/core/App/navigators/OnboardingScreens.ts
jleach: I mainly moved this code to make it easier to add, remove, or reorder existing onboarding screens. If a jurisdiction wanted to introduce a unique screen just for themselves, we’d need to do some additional refactoring. Specifically, `generateOnboardingWorkflowSteps` would need to return the actual components.  

I see this as an initial "crawl" step toward fully customizable onboarding workflows, where `generateOnboardingWorkflowSteps` would ultimately handle returning the components directly. Let me think through this change and see if it can be included.

packages/legacy/core/App/navigators/OnboardingStack.tsx
```diff
@@ -15,36 +15,37 @@ import { createCarouselStyle } from '../screens/OnboardingPages'
 import PINCreate from '../screens/PINCreate'
 import PINEnter from '../screens/PINEnter'
 import PushNotification from '../screens/PushNotification'
+// import UpdateAvailable from '../screens/UpdateAvailable'
```
fc-santos: Any reason to keep this comment?

jleach: Fixed.

packages/legacy/core/App/navigators/OnboardingStack.tsx
```diff
@@ -83,122 +91,78 @@ const OnboardingStack: React.FC = () => {
...
+    ]
+  )
   return (
-    <Stack.Navigator initialRouteName={Screens.Splash} screenOptions={{ ...defaultStackOptions }}>
+    <Stack.Navigator
+      initialRouteName={Screens.Preface}
+      screenOptions={{
+        ...defaultStackOptions,
+      }}
+    >
```
fc-santos, line 165: The Preface Screen is optional, what happens if you configured your wallet to not have a Preface ?

jleach: The only screens that are required are create pin and/or enter pin. We don't have a way to programmatically enforce this so it's just a conversation for now. If you don't add a preface you would just go to the next screen in the array until you reach a required screen.

fc-santos (dismissed):

packages/legacy/core/App/navigators/OnboardingScreens.ts
```diff
@@ -0,0 +1,140 @@
...
+interface ScreenComponents {
+  Splash: React.FC<StackScreenProps<ParamListBase>>
+  Preface: React.FC<StackScreenProps<ParamListBase>>
+  Terms: React.FC<StackScreenProps<ParamListBase>>
+  NameWallet: React.FC<StackScreenProps<ParamListBase>>
+  useBiometry: React.FC<StackScreenProps<ParamListBase>>
+  PushNotification: React.FC<StackScreenProps<ParamListBase, Screens.UsePushNotifications>>
+  Developer: React.FC<StackScreenProps<ParamListBase>>
+  AttemptLockout: React.FC<StackScreenProps<ParamListBase>>
+  OnBoardingScreen: React.FC
```
bryce-mcmath: Should we take the opportunity to change this while we're touching this code, so it's the same casing as it is everywhere else?
```suggestion
  OnboardingScreen: React.FC
```

packages/legacy/core/App/navigators/OnboardingStack.tsx
```diff
@@ -73,7 +81,7 @@ const OnboardingStack: React.FC = () => {
     [dispatch]
   )
 
-  const OnBoardingScreen = () => {
+  const OnBoardingScreen = useCallback(() => {
```
bryce-mcmath:
```suggestion
  const OnboardingScreen = useCallback(() => {
```

packages/legacy/core/__tests__/screens/Splash.test.tsx
```diff
@@ -105,6 +105,6 @@ describe('Splash Screen', () => {
       jest.runAllTimers()
     })
 
-    expect(navigation.dispatch).toHaveBeenCalledTimes(1)
+    expect(navigation.dispatch).toHaveBeenCalledTimes(0)
```
al-rosenthal: nit: what are we testing for if we don't expect this to be called?

packages/legacy/core/App/screens/Terms.tsx
```diff
@@ -16,7 +16,7 @@ import { AuthenticateStackParams, Screens } from '../types/navigators'
 import { testIdWithKey } from '../utils/testable'
 import ScreenLayout from '../layout/ScreenLayout'
 
-export const TermsVersion = true
+export const TermsVersion = '1'
```
al-rosenthal, line 19: how will this impact users that are upgrading from a boolean to a string?

packages/legacy/core/App/navigators/RootStack.tsx
```diff
@@ -20,20 +20,21 @@ const RootStack: React.FC = () => {
...
   const declinedProofs = useProofByState([ProofState.Declined, ProofState.Abandoned])
+  const [onboardingComplete, setOnboardingComplete] = useState(false)
 
   useDeepLinks()
 
-  const isAuthenticated = useMemo(
-    () => store.authentication.didAuthenticate && store.onboarding.postAuthScreens.length === 0,
-    [store.authentication.didAuthenticate, store.onboarding.postAuthScreens]
-  )
+  const isOnboardingComplete = useMemo(() => onboardingComplete, [onboardingComplete])
```
bryce-mcmath, line 27: I don't think this `useMemo` will have any effect. I think primitive values already only cause re-renders when they change. If they are reassigned to the same value they don't cause re-renders (afaik)

packages/legacy/core/App/navigators/OnboardingStack.tsx
```diff
@@ -16,35 +16,35 @@ import PINCreate from '../screens/PINCreate'
...
   const Stack = createStackNavigator()
   const theme = useTheme()
   const OnboardingTheme = theme.OnboardingTheme
   const carousel = createCarouselStyle(OnboardingTheme)
   const [
-    splash,
+    config,
+    Splash,
     pages,
     useBiometry,
```
bryce-mcmath, line 38: Since this is a screen (JSX) and not a hook it should probably be capitalized. Not something we have to do in this PR, just at some point

packages/legacy/core/App/screens/NameWallet.tsx
```diff
@@ -83,19 +83,17 @@ const NameWallet: React.FC = () => {
...
       })
+
       if (agent) {
         agent.config.label = walletName
       }
+
       dispatch({ type: DispatchAction.DID_NAME_WALLET })
-      if (store.onboarding.didCompleteOnboarding) {
-        navigation.goBack()
-      } else {
```
bryce-mcmath, line 96: I think this screen can be accessed from the settings menu, not just onboarding. So that call to `navigation.goBack()` might still be needed in that case

packages/legacy/core/App/screens/Splash.tsx
```diff
@@ -104,140 +36,9 @@ const Splash: React.FC = () => {
     },
   })
 
-  // navigation calls that occur before the screen is fully mounted will fail
-  // this useeffect prevents that race condition
-  useEffect(() => {
-    setMounted(true)
-  }, [])
-
-  useEffect(() => {
```
bryce-mcmath, line 113: With this change, what does the user see while state is being loaded for the first time on a cold start?

bryce-mcmath (commented):
Few small non-blocking suggestions and some questions

packages/legacy/core/App/screens/Terms.tsx
jleach: I think it may improve things. In the code the "TermsVersion" is treated as a [string](https://github.com/search?type=code&q=repo%3Abcgov%2Fbc-wallet-mobile+%22TermsVersion%22) and in bifold it's mixed and matched also as shown [here](https://github.com/search?type=code&q=repo%3Aopenwallet-foundation%2Fbifold-wallet+%22TermsVersion%22).

I don't think "string" | "boolean" are good options. LMK.

packages/legacy/core/App/navigators/OnboardingStack.tsx
jleach: You mean "useBiometry" if so I created a ticket with some issues I noticed along the way and this was one. I think it should be renamed to not use `use` in the name for clarity.

packages/legacy/core/App/screens/Splash.tsx
jleach: I think I may need to add this back then. It's a bit confusing but we have a splash screen imported from a NPM package and our Splash screen. I assumed there was a bug in the NPM version that is meant to show the app logo screen until the "state" is loaded at which point the app can then continue. Maybe I need to update Splash to show in two different ways, or fix the npm package to work.

packages/legacy/core/App/screens/NameWallet.tsx
jleach: Fixed.

packages/legacy/core/App/navigators/OnboardingStack.tsx
jleach: Fixed

packages/legacy/core/App/navigators/RootStack.tsx
jleach: Fixed.

packages/legacy/core/App/navigators/OnboardingScreens.ts
jleach: Fixed.

packages/legacy/core/App/screens/Terms.tsx
al-rosenthal: I agree string | boolean is not a great type definition. I was mostly concerned if a user on a previous version of the app needs to accept new terms but it looks like it's already accounted for.

packages/legacy/core/App/navigators/OnboardingScreens.ts
```diff
@@ -0,0 +1,140 @@
...
+  ScreenOptionsDictionary: ScreenOptionsType,
+  components: ScreenComponents
+): ScreenOptions[] => [
+  {
+    name: Screens.Splash,
+    component: components.Splash,
+    options: {
+      ...ScreenOptionsDictionary[Screens.Splash],
+      ...TransitionPresets.ModalFadeTransition,
+      title: t('Screens.Splash'),
```
fc-santos: Can you pass the ScreenOptionsDictionary last in the options like this for all elements in the array:

```
{
    ...TransitionPresets.ModalFadeTransition,
    title: t('Screens.Splash'),
    ...ScreenOptionsDictionary[Screens.Splash],
}
```

jleach: @fc-santos I think if I do that then "title" would get overridden if there was a default in `ScreenOptionsDictionary`. It goes first to allow defaults to be overridden. Thoughts?

fc-santos: That's the objective. It's up to other wallets to decide what they want to keep as default and what they want to override. By putting the ScreenOptionsDictionary first, if let's say a wallet has a different TransitionPresets, then it would be overridden by the defaults.

jleach: Fixed.

jleach:
@fc-santos @bryce-mcmath @al-rosenthal Ready for re-review.

packages/legacy/core/App/hooks/useOnboardingState.ts
```diff
@@ -0,0 +1,36 @@
...
+      return
+    }
+
+    const screens = generateOnboardingWorkflowSteps(store, config, termsVersion)
+    setLocalState(screens)
+  }, [store, config, termsVersion, generateOnboardingWorkflowSteps])
+
+  return {
+    localState,
+    setLocalState,
```
bryce-mcmath: Does this setter need to be exposed? It doesn't seem to be used anywhere outside of the hook. I'm also wondering if pulling a var called `localState` from a hook into a regular component might be confusing. Any problem with us calling it `onboardingState` or `localOnboardingState` or something like that?

bryce-mcmath (commented):
All looks good besides just the one comment about naming + exposing setter

packages/legacy/core/App/hooks/useOnboardingState.ts
jleach: Fixed.

bryce-mcmath (approved):

PR #1477, opened 2025-03-25
fix: added store check if the user authenticated

al-rosenthal:
# Summary of Changes

- added a check if the user is authenticated before attempting to initialize the wallet agent

N/A

# Breaking change guide

N/A

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓

### DIFF
--- a/packages/legacy/core/App/screens/Splash.tsx
+++ b/packages/legacy/core/App/screens/Splash.tsx
@@ -13,6 +13,7 @@ import useInitializeAgent from '../hooks/initialize-agent'
 import { BifoldError } from '../types/error'
 import { Stacks } from '../types/navigators'
 import { useAuth } from '../contexts/auth'
+import { useStore } from '../contexts/store'
 
 /**
  * To customize this splash screen set the background color of the
@@ -23,6 +24,7 @@ const Splash: React.FC = () => {
   const { walletSecret } = useAuth()
   const { t } = useTranslation()
   const navigation = useNavigation()
+  const [store] = useStore()
   const { ColorPallet } = useTheme()
   const { LoadingIndicator } = useAnimatedComponents()
   const { initializeAgent } = useInitializeAgent()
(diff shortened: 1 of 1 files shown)


packages/legacy/core/App/screens/Splash.tsx
```diff
@@ -42,6 +44,11 @@ const Splash: React.FC = () => {
       try {
         await (ocaBundleResolver as RemoteOCABundleResolver).checkForUpdates?.()
 
+        // User hasn't authenticated yet, no point in trying to initialize wallet agent
+        if (!store.authentication.didAuthenticate) {
```
bryce-mcmath, line 48: Will have to be added to `useEffect` dependency array

fc-santos (dismissed):
So because user has not been authenticated, the wallet secret is missing. Good catch!

fc-santos:
@al-rosenthal fix lint issues

packages/legacy/core/App/screens/Splash.tsx
al-rosenthal: done

fc-santos:
We should probably create a pre commit hook, so it runs a lint check, typecheck and run the tests before actually committing. 

@al-rosenthal @bryce-mcmath

bryce-mcmath:
> We should probably create a pre commit hook, so it runs a lint check, typecheck and run the tests before actually committing

we used to have one but I guess it's broken. I'll get it working again

al-rosenthal:
> We should probably create a pre commit hook, so it runs a lint check, typecheck and run the tests before actually committing.
> 
> @al-rosenthal @bryce-mcmath

We actually have one already in `.husky/pre-commit` it's strange it did run but use a plugin for that provides a but of UI for conventional commits so it might be doing something to skip that.

bryce-mcmath (approved):

PR #1493, opened 2025-04-02
feat: prevent screenshots

bryce-mcmath:
# Summary of Changes

This PR adds the option to enable screenshot prevention using [react-native-capture-protection](https://github.com/wn-na/react-native-capture-protection). We chose this library because it is simple, lightweight, requires minimal configuration, works on both device platforms, and can be used in Expo, should we decide to migrate to Expo eventually.

Currently it is disabled by default, but even when enabled, it only applies to one screen so far: the `ProofDetails` screen, which is what the verifier sees when their proof request is responded to. Disabling screenshots here helps to discourage collecting sent information in a non-wallet context and aide's the sender's privacy. 

The amount of code to make it work is pretty small and the JSDoc should explain the exported hook well enough if you'd like to use it elsewhere in your projects.

Here's a screenshot on iOS taken on the `ProofDetails` screen:
ios_screenshot_protected

Here's a screenshot on iOS taken on the `Home` screen in the same session:
ios_screenshot_unprotected

Here's what it looks like on Android when you try and take a screenshot on the `ProofDetails` screen:
android_screenshot_protected

# Breaking change guide

`react-native-capture-protection` is now a peer dependency of Bifold. It requires running `yarn ios:setup` after adding it 

If you'd like to enable it, simply add `preventScreenCapture: true` to your config object

# Related Issues

bcgov/bc-wallet-mobile#1103

# Pull Request Checklist

Pro Tip 🤓


packages/core/src/hooks/screen-capture.ts
```diff
@@ -0,0 +1,41 @@
...
+ * const MyScreen = () => {
+ *   usePreventScreenCapture();
+ *   return <View><Text>Secure Content</Text></View>;
+ * };
+ * ```
+ *
+ * @returns {void}
+ */
+const usePreventScreenCapture = (active: boolean = true) => {
+  const [{ preventScreenCapture }] = useServices([TOKENS.CONFIG])
```
jleach, line 28: I'm wondering if this would be better passed in as a argument to so that `usePreventScreenCapture` has less dependencies. It's probably easier to test in isolation if it has less dependencies.

al-rosenthal (approved):
📷

jleach (changes requested):

packages/core/src/hooks/screen-capture.ts
bryce-mcmath: Good point, will make another quick PR with that change

PR #1507, opened 2025-04-16
chore: refactor PINEnter screen

LippaC-OPS:
### Summary of Changes
Refactor PIN Enter screen. Split it up into two screens one to Verify PIN within the app and one to unlock the app when its opened.

### Breaking change guide

N/A

### Related Issues

### Pull Request Checklist
Tick all boxes below to demonstrate that you have completed the respective task. If the item does not apply to your this PR check it anyway to make it apparent that there's nothing to do.

- [x]  All commits contain a DCO Signed-off-by line (we use the [DCO GitHub app](https://github.com/apps/dco) to enforce this)
- [x]  If applicable, screenshots, gifs, or video are included for UI changes
- [x]  If applicable, breaking changes are described above along with how to address them
- [x]  Updated documentation as needed for changed code and new or modified features
- [x]  Added sufficient [tests](https://github.com/openwallet-foundation/bifold-wallet/packages/legacy/core/__tests__/) so that overall code coverage is not reduced

If you have any questions to any of the points above, just submit and ask! This checklist is here to help you, not to deter you from contributing!

Pro Tip 🤓

Read our [contribution guide](https://github.com/openwallet-foundation/bifold-wallet/CONTRIBUTING.md) at least once; it will save you a few review cycles!
Your PR will likely not be reviewed until all the above boxes are checked and all automated checks have passed


packages/core/src/hooks/developer-mode.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const TOUCH_COUNT_TO_ENABLE_DEVELOPER_MODE = 10
+
+export const useDeveloperMode = () => {
+  const developerOptionCount = useRef(0)
+  const [, dispatch] = useStore()
+  const { navigate } = useNavigation()
+
+  const incrementDeveloperMenuCounter = useCallback(() => {
+    if (developerOptionCount.current === TOUCH_COUNT_TO_ENABLE_DEVELOPER_MODE) {
```
al-rosenthal: I think I'd change the check to: 
```typescript
if (developerOptionCount.current >= TOUCH_COUNT_TO_ENABLE_DEVELOPER_MODE) {
...
```

I'm worried that it might count properly but could potentially be missed if the count needs to be exactly 10

packages/core/src/screens/PINVerify.tsx
```diff
@@ -0,0 +1,177 @@
...
+  const { verifyPIN } = useAuth()
+  const [PIN, setPIN] = useState<string>('')
+  const [continueDisabled, setContinueDisabled] = useState(false)
+  const [loading, setLoading] = useState(false)
+  const [alertModalVisible, setAlertModalVisible] = useState<boolean>(false)
+  const { ColorPallet, TextTheme } = useTheme()
+  const { ButtonLoading } = useAnimatedComponents()
+  const [inlineMessageField, setInlineMessageField] = useState<InlineMessageProps>()
+  // Temporary until all use cases are built with the new design
+  const isNewDesign = usage === PINEntryUsage.ChangeBiometrics
```
al-rosenthal, line 39: What is this new design? is it something coming in another PR?

packages/core/src/screens/PINEnter.tsx
```diff
@@ -438,109 +215,57 @@ const PINEnter: React.FC<PINEnterProps> = ({ setAuthenticated, usage = PINEntryU
...
-          )}
-          <ThemedText variant="bold" style={style.subText}>
-            {inputLabelText[usage]}
-            {usage === PINEntryUsage.ChangeBiometrics && (
-              <ThemedText style={style.parenthesisText}>
-                {' '}
-                {t('PINEnter.ChangeBiometricsInputLabelParenthesis')}
-              </ThemedText>
-            )}
+      <View style={[style.screenContainer, { backgroundColor: ColorPallet.brand.primaryBackground }]}>
```
al-rosenthal: nit: this could be moved into the style sheet

al-rosenthal (commented):
couple of small nits and some sonar cloud issues we should clean up but otherwise great refactor

bryce-mcmath:
This is kind of a larger conversation but do we actually need the postAuthScreens stuff now that we have the new onboarding workflow generation?

packages/core/src/screens/PINEnter.tsx
LippaC-OPS: I had to move the stylesheet out of the component because otherwise it was pointless to have, and couldn't access the ColorPallet outside the component since it comes from a hook so I combined them here. 

Is there an alternative way to access the ColorPallet not through a hook?

packages/core/src/hooks/developer-mode.ts
LippaC-OPS: Sure, I can make this change no problem

packages/core/src/screens/PINVerify.tsx
LippaC-OPS: So we started implementing a new design on these screens but realized the refactor here should be done first. So once this is merged I can add the new design to the PINCheck variant of this screen as well and remove this.

packages/core/src/screens/PINEnter.tsx
al-rosenthal: you could nest the stylesheet inside of the PINEnter component, then it would have access to the color or theme hooks. That pattern is also consistent with a lot of other components in bifold

LippaC-OPS: The reason the stylesheet is supposed to be outside the component is so that its not recreated on every render. Having the stylesheet within the component is not really best practice. Although its only a minimal performance impact.

al-rosenthal: that's a good point, ok well let's leave it as is then. This is probably a good topic for technical debt/ clean up

packages/core/src/screens/PINVerify.tsx
```diff
@@ -0,0 +1,177 @@
...
+    width: '100%',
+  },
+  helpText: {
+    alignSelf: 'auto',
+    textAlign: 'left',
+  },
+  modalText: {
+    marginVertical: 5,
+  },
+})
```
bryce-mcmath: I know there's a minor performance benefit to keeping the stylesheets out of the components, but in order to incorporate theme (which comes from a hook) into stylesheets and keep it reactive, I think we should continue to follow the same pattern we have throughout the rest of Bifold. Also worth noting that we're not going to see the performance benefit if we still include a bunch of inline styles as well (as we are in this component, for example)

packages/core/src/screens/PINEnter.tsx
```diff
@@ -625,4 +337,29 @@ const PINEnter: React.FC<PINEnterProps> = ({ setAuthenticated, usage = PINEntryU
...
+  inputLabel: {
+    marginBottom: 16,
+  },
+  modalText: {
+    marginVertical: 5,
+  },
+  subTitle: {
+    marginBottom: 20,
+  },
+})
```
bryce-mcmath: "

bryce-mcmath (changes requested):
Lots of good refactoring here 🙌 One issue: I'm not sure moving stylesheets outside of components is the right approach when nearly every component uses a reactive theme that comes from a hook. Any performance benefit we get from moving them out will be negated by all the inlines styles we'll have to add afterwards to keep the theming. Let me know what you think, your other changes look great

packages/core/src/screens/PINVerify.tsx
LippaC-OPS: I completely understand and I have no issues with moving it back inside the component if that is the general consensus. I'll make that change and push it shortly.

MosCD3 (dismissed):

bryce-mcmath:
@LippaC-OPS looks good! You can add a patch changset with `yarn changeset add`

That will automatically update our changelog and get a new release ready. You can refer to Mostafa's PR as an example or a lot of the recent closed ones :)

bryce-mcmath (approved):

PR #1479, opened 2025-03-26
refactor: Apply field attribute names mapping for OpenID credentials

MosCD3:
# Summary of Changes

Replace this text with a high-level summary of the changes included in this PR.

# Screenshots, videos, or gifs

Replace this text with embedded media for UI changes if they are included in this PR. If there are none, simply enter N/A

# Breaking change guide

Replace this text with any breaking changes included in this PR along with how to address them in downstream projects. If there are none, simply enter N/A

# Related Issues

Replace this text with issue #'s that are relevant to this PR. If there are none, simply enter N/A

# Pull Request Checklist

Tick all boxes below to demonstrate that you have completed the respective task. If the item does not apply to your this PR **check it anyway** to make it apparent that there's nothing to do.

- [ ] All commits contain a DCO `Signed-off-by` line (we use the [DCO GitHub app](https://github.com/apps/dco) to enforce this)
- [ ] If applicable, screenshots, gifs, or video are included for UI changes
- [ ] If applicable, breaking changes are described above along with how to address them
- [ ] Updated documentation as needed for changed code and new or modified features
- [ ] Added sufficient [tests](../packages/legacy/core/__tests__/) so that overall code coverage is not reduced

If you have _any_ questions to _any_ of the points above, just **submit and ask**! This checklist is here to _help_ you, not to deter you from contributing!

Pro Tip 🤓

- Read our [contribution guide](../CONTRIBUTING.md) at least once; it will save you a few review cycles!
- Your PR will likely not be reviewed until all the above boxes are checked and all automated checks have passed

PR #1508, opened 2025-04-21
fix(ui): re-create OpenID UI fixes

MosCD3:
Created instead of #1479 

1- Add missing credential_subject to metadata
2- Re create the fields extraction function to account for substituting any field title from credential_subject fields
3- Other minor theme changes and localization


packages/core/src/utils/oca.ts
```diff
@@ -11,32 +11,73 @@ import {
...
+
+export const buildFieldsFromW3cCredsCredential = (
+  display: W3cCredentialDisplay,
+  filterByAttributes?: string[]
+): Array<Field> => {
+  return (
+    Object.entries(display.attributes)
+      .filter(([key]) => key !== 'id')
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      .map(([key, value]) => getAttributeField(display, key))
```
bryce-mcmath:
```suggestion
      .map(([key]) => getAttributeField(display, key))
```

packages/core/src/utils/oca.ts
```diff
@@ -11,32 +11,73 @@ import {
...
+  }
+
+  //Now check credentialSubject for attributeName mapping
+  const credentialSubject = display.credentialSubject
+
+  if (credentialSubject) {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    for (const [key, value] of Object.entries(credentialSubject)) {
+      if (key !== searchKey) continue
+      if (!value) continue
```
bryce-mcmath: You're using both variables so you can safely remove the eslint exception, no? Can also merge those two conditionals
```suggestion
    for (const [key, value] of Object.entries(credentialSubject)) {
      if (key !== searchKey || !value) continue
```

bryce-mcmath (changes requested):
Some minor things, also try to address some of the SonarCloud issues if you can

bryce-mcmath:
Nice work! Try adding a patch changeset with `yarn changeset add` :)

.changeset/tasty-windows-stick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@bifold/core': major
```
bryce-mcmath: nit: I think this should just be `patch` rather than `major`, since you are only fixing bugs not making breaking changes (major) or adding new features (minor)

packages/core/src/utils/oca.ts
```diff
@@ -11,32 +11,73 @@ import {
 import { W3cCredentialDisplay } from '../modules/openid/types'
 import { BrandingOverlay } from '@bifold/oca'
 
-export const buildFieldsFromAnonCredsCredential = (credential: CredentialExchangeRecord): Array<Field> => {
-  return credential?.credentialAttributes?.map((attr) => new Attribute(attr)) || []
+export type FiledExt = {
```
bryce-mcmath: Just a question, was this meant to be FieldExt rather than FiledExt? I'm not super familiar with the OpenID stuff

bryce-mcmath (dismissed):
Just a small nit about the versioning chosen (major -> breaking change, minor -> new feature, patch -> bug fixes)

ref: https://semver.org/

.changeset/tasty-windows-stick.md
MosCD3: Can you guide me through this :). Its new to me

packages/core/src/utils/oca.ts
MosCD3: Ops, typ

MosCD3: It supposed to be Field Extended so FieldExt

bryce-mcmath (approved):
Noice!

PR #1512, opened 2025-04-25
feat: proper agentprovider use and configurable theme

bryce-mcmath:
(Our PR template is broken currently, @al-rosenthal is addressing that in his open PR)

This PR does two main things that are closely related:
- enables having multiple themes, switching between them
- refactor to properly use @credo-ts/react-hooks, mainly the AgentProvider (no longer calling setAgent from within the provider). This removes the need for the very large and ugly patch we'd added and fixes some outstanding bugs with websockets and agent setup / shutdown

There is also a bit of cleanup in this PR, notably:
- splitting out the `PINCreate` screen into `PINCreate` and `PINChange` screens that are easier to read and require only what is needed for specific usage (no agent required in `PINCreate`)
- similar to above, split out `PushNotification` to `PushNotifications` (onboarding, using same name as screen title) and `TogglePushNotifications` (settings)
- similar to above, split out `UseBiometry` screen into `Biometry` screen (onboarding) and `ToggleBiometry` screen (settings)
- made common components for the content of the above screens to prevent duplication
- no more `...wrapped in act(...)` errors when Jest tests are run
- no more duplicate ariesbifold / AriesBifold Xcode workspace
- Navigation container component with reactive theme
- Rotation properly disabled on iPhones, was previously only prevented on Android 
- added a .tools-versions for easy environment setup with things like `mise` (https://mise.jdx.dev/)


bryce-mcmath:
> 20 new sonarcloud issues

These all existed already but are being triggered from renamed files. I fixed the one new issue that arose

al-rosenthal (approved):
ran this locally on both ios and android and it all seems good to me, great job!

PR #1570, opened 2025-06-26
docs: Explain openJDK dependency and how to manually install

ehanoc:
# Summary of Changes

The instructions on linux weren't too clear on what specific version of JDK is needed. The build step seems to be coupled to that specific version and to the `zulu` version. 

This adds a bit more information about that and explains how to manually install in Linux distros that don't come with that version available in the package manager. Which was my case.


al-rosenthal (approved):
thanks for the update!

bryce-mcmath (approved):

PR #1594, opened 2025-07-25
chore: logging with stack trace

Mkg0092:
# Summary of Changes

PR to add stack trace to the log output of logger.error and logger.fatal.

Screenshot 2025-07-25 at 3 05 49 PM

# Breaking change guide

 N/A

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


packages/remote-logs/src/logger.ts
```diff
@@ -129,32 +129,34 @@ export class RemoteLogger extends BifoldLogger {
...
-  public warn(message: string, data?: object | undefined): void {
-    this._log?.warn({ message, data })
+  public warn(...msgs: unknown[]): void {
+    this._log?.warn(...msgs)
   }
 
-  public error(message: string, data?: object | undefined): void {
-    this._log?.error({ message, data })
+  public error(...msgs: unknown[]): void {
+    const stack = new Error().stack ?? ''
```
cvarjao: this logs where the logging is being called, not where the error got raised. The original error/stack is still being hidden/lost

bryce-mcmath: @MKGpp you could try checking for the first instance of an `Error` object (BifoldError and CredoError both extend the base Error object) in `msgs` and log that error's stack if it exists. That might be the easiest way to add the proper stacktrace without having to make too many changes right away. Then we can go through and make sure when we call `logger.error(...)`, we're passing an Error object of some kind.

bryce-mcmath: We do have a bit of bad pattern of creating `new BifoldError`s and only supplying the message property from the original error, losing the more accurate stacktrace (eg. L114 of RootStack.tsx) but the above would still be an improvement in the interim. Not sure if we want to address fixing that pattern now while we're working on this - the pattern is pretty widespread

packages/remote-logs/src/logger.ts
```diff
@@ -129,32 +129,34 @@ export class RemoteLogger extends BifoldLogger {
...
 
-  public info(message: string, data?: object | undefined): void {
-    this._log?.info({ message, data })
+  public info(...msgs: unknown[]): void {
+    this._log?.info(...msgs)
   }
 
-  public warn(message: string, data?: object | undefined): void {
-    this._log?.warn({ message, data })
+  public warn(...msgs: unknown[]): void {
```
MacDeluca, line 148: suggestion: I would also check to see if the log output looks the same from the original version to your updated one. 
ie: ({message, data}) => (message, data)
My assumption is the output will look different now.

Also, a simple fix for now could be adding an Error as the known types for data. ie: object | Error | undefined. Then you could check if (data instanceof Error) and grab the properties needed, or just log the error as is.

And optionally, if we still wanted the data alongside the error you could do: 

```ts
(message: string, data?: object | Error, error?: Error)
```

Which would grant us some strictness with the typing, while also allowing some flexibility with logs having Errors + the data object metadata.

```ts
logger.error('This is an error log')
logger.error('This is an error log with an error', error)
logger.error('This is an error log with meta and error', metadata, error)

```

bryce-mcmath (approved):

bryce-mcmath:
🪳🔨

PR #1619, opened 2025-08-20
feat: pin flow nist compliance

LippaC-OPS:
# Summary of Changes

PIN flow changes with a new design, added handling for NIST compliance specifications. All old PIN flows are intact and can be swapped within the app config.

Change PIN flow
https://github.com/user-attachments/assets/c46781f6-922c-43f6-ab83-42590e334ca9

Create PIN flow
https://github.com/user-attachments/assets/c34388b8-eac6-484d-aeb2-78ce37705a4e

Enter PIN flow
https://github.com/user-attachments/assets/d63f6e03-ba09-4b5f-a847-be93d95d46bf

# Breaking change guide

# Related Issues

# Pull Request Checklist

Pro Tip 🤓


MosCD3 (dismissed):

MosCD3:
@LippaC-OPS can you resolve conflicts here

packages/core/__tests__/screens/__snapshots__/PINEnter.test.tsx.snap
```diff
@@ -37,6 +37,24 @@ exports[`PINEnter Screen PIN Enter renders correctly 1`] = `
...
+                },
+                Object {
+                  "marginBottom": 24,
+                  "marginTop": 16,
+                },
+              ]
+            }
+          >
+            PINEnter.AppTitle
+          </Text>
```
bryce-mcmath: If this is feature flagged, should this new text be showing up in the snapshot? Assuming the snapshot test is wrapped in the BasicAppContext and that context is configured to be the same as before, then nothing should really change in snapshots right?

packages/core/src/components/inputs/PINInput.tsx
```diff
@@ -21,29 +23,42 @@ interface PINInputProps {
...
 ) => {
+  const [{ PINScreensConfig }] = useServices([TOKENS.CONFIG])
+
+  const { PINInputTheme, SeparatedPINInputTheme, ColorPalette } = useTheme()
+
+  const theme = PINScreensConfig.useNewPINDesign ? SeparatedPINInputTheme : PINInputTheme
+
   const [PIN, setPIN] = useState('')
-  const [showPIN, setShowPIN] = useState(false)
+  const [showPIN, setShowPIN] = useState(PINScreensConfig.useNewPINDesign ? true : false)
```
jleach: Could that be null or undefined? I wonder if `const [showPIN, setShowPIN] = useState(PINScreensConfig.useNewPINDesign ?? false)` would be better?

bryce-mcmath (changes requested):
Couple small general changes:

Can we conform with the rest of the project wrt stylesheets being kept inside the component? That way it's easy to add reactive properties (like theme properties) to them. I think the performance gain is negligible when weighed against the ability to reactively style and keep the codebase consistent

If it's feature flagged and off by default, there shouldn't be changes to existing snapshots (other than the ones changing slightly due to now being wrapped with the BasicAppContext, those are fine). But new text elements showing up probably shouldn't happen

packages/core/src/components/misc/PINScreenTitleText.tsx
```diff
@@ -0,0 +1,34 @@
...
+        {header}
+      </ThemedText>
+      <ThemedText variant="headerTitle" style={style.subheader}>
+        {subheader}
+      </ThemedText>
+    </View>
+  )
+}
+
+const style = StyleSheet.create({
```
jleach: We don't really use this format throughout the app. Can we make it like the other screens where the style is defined inside the comonent?

packages/core/src/components/modals/ConfirmPINModal.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { InlineMessageProps } from '../../components/inputs/InlineErrorText'
+import PINInput from '../../components/inputs/PINInput'
+import usePreventScreenCapture from '../../hooks/screen-capture'
+import { usePINValidation } from '../../hooks/usePINValidation'
+import { useTheme } from '../../contexts/theme'
+import { TOKENS, useServices } from '../../container-api'
+import { testIdWithKey } from '../../utils/testable'
+
+interface ConfirmPINModalProps {
+  errorMessage: InlineMessageProps | undefined
```
jleach: Would `errorMessage?: InlineMessageProps` be the same but more clear?

packages/core/src/container-impl.ts
```diff
@@ -84,6 +84,10 @@ export const defaultConfig: Config = {
     appleAppStoreUrl: 'https://example.com',
     googlePlayStoreUrl: 'https://example.com',
   },
+  PINScreensConfig: {
+    useNewPINDesign: false,
+    nistPINRequirements: false,
```
al-rosenthal: it looks like `nistPINRequirements` isn't being used, is this just setup for something else later?

packages/core/src/screens/PINEnter.tsx
```diff
@@ -364,38 +380,54 @@ const PINEnter: React.FC<PINEnterProps> = ({ setAuthenticated }) => {
...
-              accessibilityLabel={t('PINEnter.Unlock')}
-              onPress={() => {
-                Keyboard.dismiss()
-                onPINInputCompleted(PIN)
-              }}
-            >
-              {!continueEnabled && <ButtonLoading />}
-            </Button>
-          </View>
+          {!PINScreensConfig.useNewPINDesign && (
```
al-rosenthal, line 437: No button needed for the this screen in the new mode? This button also has a loading indicator which gives the user some context that something is happening. On older devices (tested with an iPhone 7) it feels like nothing happens once the pin is entered and begins to open with wallet

packages/core/src/screens/PINVerify.tsx
```diff
@@ -46,46 +49,59 @@ const PINVerify: React.FC<Props> = ({ setAuthenticated, usage = PINEntryUsage.PI
...
-      setAlertModalVisible(true)
-    }
-    setLoading(false)
-    setContinueDisabled(false)
-  }, [verifyPIN, setLoading, setAuthenticated, setContinueDisabled, PIN])
+    usage !== PINEntryUsage.ChangePIN && setAuthenticated(false)
+  }, [setAlertModalVisible, setAuthenticated, usage])
+
+  const onPINInputCompleted = useCallback(
+    async (p?: string) => {
```
jleach: Can we use a more descriptive variable name like `userPinInput` it just makes reading code easier.

jleach (changes requested):

packages/core/src/screens/PINCreate.tsx
```diff
@@ -116,13 +150,17 @@ const PINCreate: React.FC<PINCreateProps> = ({ setAuthenticated, explainedStatus
   return explained ? (
     <KeyboardView keyboardAvoiding={false}>
```
al-rosenthal: The new input is completely covered by the keyboard on both devices I tested with.


### PIN Create on an iPhone 7 
IMG_0008

### PIN Create on an OnePlus10
dummy

al-rosenthal (changes requested):
Ran this locally on both iOS and android and overall it works well. I did run into an issue with the change PIN:
1. Finish onboarding
2. Navigate to PIN change
3. Enter old PIN 
4. Enter new PIN (221133)
5. Enter new PIN again (332244)
6. New PIN is saved and the first new PIN (221133) can be used to unlock the wallet

packages/core/__tests__/screens/__snapshots__/PINEnter.test.tsx.snap
LippaC-OPS: @bryce-mcmath There is some slight changes in how the title is displayed at the top of the screen. If you would like me to put that behind the feature flag as well let me know

packages/core/src/components/inputs/PINInput.tsx
LippaC-OPS: sure no problem

packages/core/src/components/misc/PINScreenTitleText.tsx
LippaC-OPS: Will make this change no worries, just habitually put it outside of the component but no issues putting it inside 👍

packages/core/src/components/modals/ConfirmPINModal.tsx
LippaC-OPS: No problem will make this change

packages/core/src/screens/PINVerify.tsx
LippaC-OPS: Changed all the p variables to this to be more descriptive 👍

packages/core/src/container-impl.ts
LippaC-OPS: This was meant to be used as a shortcut config if people wanted to use the nist pin requirements so length is 6, and the only other limitation is that its not a commonly used pin with the list provided. Must have been moved around in testing, good catch thanks

packages/core/src/screens/PINEnter.tsx
LippaC-OPS: Adding a loading indicator, but as per the designs i received there is no button just when user finishes inputting the pin to the required length it will automatically call the submit funciton

packages/core/src/screens/PINCreate.tsx
LippaC-OPS: Changed keyboard avoiding property to true on the parent views in these screens, good catch thanks

LippaC-OPS:
> Ran this locally on both iOS and android and overall it works well. I did run into an issue with the change PIN:
> 
> 1. Finish onboarding
> 2. Navigate to PIN change
> 3. Enter old PIN
> 4. Enter new PIN (221133)
> 5. Enter new PIN again (332244)
> 6. New PIN is saved and the first new PIN (221133) can be used to unlock the wallet

@al-rosenthal Good catch, just working on solving this issue and I will push all the PR changes, thanks!

packages/core/__tests__/screens/__snapshots__/PINEnter.test.tsx.snap
bryce-mcmath: If it's not too much work then yes please! Thanks Christian

packages/core/src/screens/PINEnter.tsx
timbl-ont: Just to confirm there is no button planned. We do need a loading indicator.

bryce-mcmath:
@LippaC-OPS since this change should be unnoticeable for teams not opting in to the new nist option, can it be a patch changeset rather than a minor? we don't really follow semantic versioning rules due to the nature of this project and it would be good to keep the major and minor versions as low as possible. otherwise every couple months the minor version will have jumped by like 20.

LippaC-OPS:
> @LippaC-OPS since this change should be unnoticeable for teams not opting in to the new nist option, can it be a patch changeset rather than a minor? we don't really follow semantic versioning rules due to the nature of this project and it would be good to keep the major and minor versions as low as possible. otherwise every couple months the minor version will have jumped by like 20.

@bryce-mcmath no worries, i will change it

packages/core/__tests__/screens/__snapshots__/PINEnter.test.tsx.snap
LippaC-OPS: @bryce-mcmath I was able to hide all the visual changes behind the feature flag but there are still changes to the snapshots. Is there someway we can move forward with the changes to the snapshots although the visuals are identical to the main branch? just pushing those changes now.

packages/core/src/screens/PINCreate.tsx
```diff
@@ -87,42 +109,58 @@ const PINCreate: React.FC<PINCreateProps> = ({ setAuthenticated, explainedStatus
...
+    return isLoading || PIN.length < minPINLength
+  }, [isLoading, PIN, inlineMessages])
 
   const continueCreatePIN = useCallback(() => {
     setExplained(true)
   }, [])
 
   return explained ? (
-    <KeyboardView keyboardAvoiding={false}>
+    <KeyboardView keyboardAvoiding={true}>
```
al-rosenthal, line 151: I tried this out again on an iPhone 7 and it scrolls when entering the first PIN but when I get to the 'Re-enter your PIN...' screen it no longer scrolls

al-rosenthal (dismissed):
I ran this locally on both android and an iPhone and couldn't get past the onboarding for a new account. It always fails on the re-enter PIN step, the PINs never match.

There are also a few sonar cloud issues that have been introduced that should be resolved before merging this

packages/core/src/screens/PINCreate.tsx
LippaC-OPS: This is fixed now I added a keyboard avoiding view to the confirm pin screen.

LippaC-OPS:
> I ran this locally on both android and an iPhone and couldn't get past the onboarding for a new account. It always fails on the re-enter PIN step, the PINs never match.
> 
> There are also a few sonar cloud issues that have been introduced that should be resolved before merging this

This should be fixed now with the most recent changes

bryce-mcmath (approved):
Tested, it works as before with just a couple new translation entries needed for new text that gets added even if a team hasn't opted in to the new flow. Not a blocker though, approved

jleach (approved):

PR #1631, opened 2025-09-03
fix: fixed react imports, package, setup and missing export

al-rosenthal:
# Summary of Changes

- add react imports in .tsx files that were missing it
- added missing export for the camera torch
- modified package json to properly build release bundles for modules

n/a

# Breaking change guide

Any components still importing bifold components like this `bifold/core/src/some/component` will break with these changes

# Related Issues

n/a

# Pull Request Checklist

Pro Tip 🤓


packages/core/src/components/misc/CredentialCard11Logo.tsx
```diff
@@ -1,12 +1,13 @@
 import { BrandingOverlay } from '@bifold/oca'
 import { BrandingOverlayType, CredentialOverlay } from '@bifold/oca/build/legacy'
+import React from 'react'
```
bryce-mcmath, line 3: Why are these required now?

al-rosenthal: During compilation/ build of JSX it will try and convert `React.createElement()` -> `_react.createElement()` but because react isn't explicitly imported it will fail at load time because `React is missing`. So in your JSX files you'd have un compiled functions like `React.createElement(...)`

This is an issue now because the source files would keep react in memory

LippaC-OPS: I am a bit confused here I thought after React 17 this wasn't necessary

packages/oca/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "TypeScript implementation of Overlay Capture Architecture (OCA) for styling Aries Verifiable Credentials",
   "main": "build/index.js",
   "types": "build/index.d.ts",
-  "source": "src/index.ts",
+  "source": "build/index.js",
```
jleach, line 7: How come this one does not get the `commonjs` path like in `build/commonjs/index.js`, for example, like react-native-attestation does?

al-rosenthal: It's how the OCA module builds, the output folder doesn't have that same structure as the other modules. Looks like it might be using a different build command, `yarn run compile` vs `bob build` in the core module

packages/core/src/components/misc/CredentialCard11Logo.tsx
al-rosenthal: ooh, I did not know that. The tsconfig will need to be updated to account for that

al-rosenthal: So that looks like it is the case for react, but  not for react native. After trying a few different configuration options in the tsconfig and babel.config it was still complaining about react missing. So I think for now we'll leave the imports

bryce-mcmath (approved):

PR #1664, opened 2025-09-23
feat: remove sub field in sd jwt credential display

LippaC-OPS:
# Summary of Changes

Filter out sub field in sd jwt credential display to hide it from the user

N/A

# Breaking change guide

N/A

# Related Issues

N/A

# Pull Request Checklist

Pro Tip 🤓


bryce-mcmath:
@jleach do you recall there being an issue with credentials that use an "id" field because of this code? Do you think this will now cause the same issue with credentials that contain a "sub" field? Wouldn't that include attestation credentials typically?

jleach:
> @jleach do you recall there being an issue with credentials that use an "id" field because of this code? Do you think this will now cause the same issue with credentials that contain a "sub" field? Wouldn't that include attestation credentials typically?

@bryce-mcmath Yes.  I think when you use `id` it conflicts with something in Credo. We noticed this while testing `did:web` or `did:webvh` or whatever it is called now. They created a credential that used `id` as a property and the wallet would not accept it. I tracked it down to a limitation in Credo.

bryce-mcmath:
@LippaC-OPS looks good, do you want to add a patch changeset with `yarn changeset add`?

LippaC-OPS:
@bryce-mcmath np, changeset added 👍

packages/core/src/utils/oca.ts
```diff
@@ -66,7 +66,7 @@ export const buildFieldsFromW3cCredsCredential = (
 ): Array<Field> => {
   return (
     Object.entries(display.attributes)
-      .filter(([key]) => key !== 'id')
+      .filter(([key]) => key !== 'id' && key !== 'sub')
```
timbl-ont: Can you also filter the Status attribute as well. 

In the future we will need to figure out how the user can access these attributes in some sort of advanced view, especially important for the sub. One option is the address book linked to the issuer.

LippaC-OPS: Status attribute is now being filtered 👍

MosCD3 (approved):

PR #1635, opened 2025-09-04
feat: refactor UI to decouple from Credential types

MosCD3:
# Summary of Changes

This PR refactors credential card rendering to decouple the UI components from OCA bundle logic and unify data handling across credential types. Key updates include:
	•	Introduced a mapper-first architecture:
	•	mapAnonCredsToCard and related mappers now handle attribute selection, primary/secondary ordering, PII detection (allPI), and proof request filtering.
	•	UI components (Card11Pure, Card10Pure) now only render the mapped data without interpreting OCA overlays.
	•	Added proof request support:
	•	When proofContext is true and displayItems are passed, the mapper uses only those attributes/predicates in the given order.
	•	Prevents leakage of extra credential attributes in proof cards.
	•	Updated CredentialCardGen.tsx to pass displayItems through to the mapper during proof flows.
	•	Removed redundant attribute reordering logic from UI components (e.g. Card11Pure).
	•	Ensured allPI and status are computed in the mapper and exposed in the unified card data model.
	
The work is still in progress

Fixes #1437

# Screenshots, videos, or gifs

Credential Rendering - Frame 1

# Breaking change guide

	•	Breaking: UI components no longer reorder attributes based on primaryAttributeKey/secondaryAttributeKey.
	•	If downstream code depended on that behavior, use the new mappers to prepare data with the desired ordering.
	•	Otherwise, existing card usage remains unchanged.
# Related Issues

Replace this text with issue #'s that are relevant to this PR. If there are none, simply enter N/A

# Pull Request Checklist

Tick all boxes below to demonstrate that you have completed the respective task. If the item does not apply to your this PR **check it anyway** to make it apparent that there's nothing to do.

- [ ] All commits contain a DCO `Signed-off-by` line (we use the [DCO GitHub app](https://github.com/apps/dco) to enforce this)
- [ ] If applicable, screenshots, gifs, or video are included for UI changes
- [ ] If applicable, breaking changes are described above along with how to address them
- [ ] If applicable, added [changeset(s)](https://github.com/changesets/changesets)
- [ ] Added sufficient [tests](../packages/core/__tests__/) so that overall code coverage is not reduced

If you have _any_ questions to _any_ of the points above, just **submit and ask**! This checklist is here to _help_ you, not to deter you from contributing!

Pro Tip 🤓

- Read our [contribution guide](../CONTRIBUTING.md) at least once; it will save you a few review cycles!
- Your PR will likely not be reviewed until all the above boxes are checked and all automated checks have passed

PR #1667, opened 2025-09-26
feat: enhance CredentialCard and related components to resolve creden…

brianorwhatever:
…tial identifiers and names using agent context

- Integrated useAgent hook to fetch credential definition ID, schema ID, and credential name dynamically.
- Updated CredentialCard, CredentialCard10, CredentialDetails, and CredentialOffer components to utilize resolved identifiers and names.
- Improved handling of credential data retrieval for better user experience and consistency across components.

### DIFF
--- a/packages/core/src/components/misc/CredentialCard.tsx
+++ b/packages/core/src/components/misc/CredentialCard.tsx
@@ -1,4 +1,5 @@
 import { CredentialExchangeRecord, MdocRecord, SdJwtVcRecord, W3cCredentialRecord } from '@credo-ts/core'
+import { useAgent } from '@credo-ts/react-hooks'
 import { Attribute, BrandingOverlayType, CredentialOverlay, Predicate } from '@bifold/oca/build/legacy'
 import React, { useEffect, useState } from 'react'
 import { ViewStyle } from 'react-native'
@@ -14,6 +15,8 @@ import { BrandingOverlay } from '@bifold/oca'
 import { useOpenIDCredentials } from '../../modules/openid/context/OpenIDCredentialRecordProvider'
 import { getCredentialForDisplay } from '../../modules/openid/display'
 import { getAttributeField } from '../../utils/oca'
+import { getCredentialIdentifiers } from '../../utils/credential'
+import { getCredentialName } from '../../utils/cred-def'
 
 interface CredentialCardProps {
   credential?: GenericCredentialExchangeRecord
@@ -50,6 +53,10 @@ const CredentialCard: React.FC<CredentialCardProps> = ({
   const [overlay, setOverlay] = useState<CredentialOverlay<BrandingOverlay>>({})
   const { resolveBundleForCredential } = useOpenIDCredentials()
   const [extraOverlayAttribute, setExtraOverlayAttribute] = useState<Attribute | undefined>()
+  const { agent } = useAgent()
+  const [resolvedCredDefId, setResolvedCredDefId] = useState<string | undefined>(credDefId)
+  const [resolvedSchemaId, setResolvedSchemaId] = useState<string | undefined>(schemaId)
+  const [resolvedCredName, setResolvedCredName] = useState<string | undefined>(credName)
 
   useEffect(() => {
     if (brandingOverlay) {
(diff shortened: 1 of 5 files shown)


bryce-mcmath:
@brianorwhatever what is the end result you're looking for here ie. what's the change in behavior before vs after this PR? Asking cause there might be a better way to do it

brianorwhatever:
> @brianorwhatever what is the end result you're looking for here ie. what's the change in behavior before vs after this PR? Asking cause there might be a better way to do it

I am ensuring all of the views of the credential I see are resolving the possible credential names. All ears for better solutions as this one has a lag where the view displays the default credential name 😅

brianorwhatever:
@bryce-mcmath do you have an alternative approach I can use? It would be great to have these resources cached so they aren't having to be downloaded for ever credential repeatedly

packages/core/src/components/misc/CredentialCard10.tsx
```diff
@@ -143,23 +146,33 @@ const CredentialCard10: React.FC<CredentialCard10Props> = ({ credential, style =
...
-    bundleResolver.resolveAllBundles(params).then((bundle) => {
+    (async () => {
+      const identifiers = getCredentialIdentifiers(credential)
+      const attributes = buildFieldsFromAnonCredsCredential(credential)
+      const credName = await getCredentialName(
+        identifiers.credentialDefinitionId,
+        identifiers.schemaId,
+        agent
+      )
+      agent?.config?.logger?.info('credName!! ' + credName)
```
brianorwhatever:
```suggestion
```

al-rosenthal:
> > @brianorwhatever what is the end result you're looking for here ie. what's the change in behavior before vs after this PR? Asking cause there might be a better way to do it
> 
> I am ensuring all of the views of the credential I see are resolving the possible credential names. All ears for better solutions as this one has a lag where the view displays the default credential name 😅

A suggestion would be to wait until the ID's and name are resolved before showing the credential. I've run this locally to test it out and haven't seen any lag you mentioned

brianorwhatever:
@al-rosenthal you will need a credential issues from a `did:webvh` issuer and uses a `did:webvh` based credential definition/schema to see the lag. Regular indy credentials are grabbing a name from within the cred def/schema `id` whereas with `did:webvh` the resource needs to be resolved before a name can be grabbed. I worry that suspending rendering until name has been retrieved might become quite slow if many credentials are in a wallet but I do think that's the right approach. Caching of the cred def/schema might be required in that case

al-rosenthal:
@brianorwhatever Ah I see. Hmm caching those values might be the way to go then with a wallet full of `did:webvh` credentials.

PatStLouis:
@al-rosenthal I think the schema should always be cached when storing a credential. While historically a credential label could be derived from the schemaId url, this is a very indy specific feature. The pattern of caching the schema as part of the credential's metadata would be an agnostic method.

MosCD3:
would it be better to merge this before -> #1635 
Remember, moving to credo-ts 6.0+ we will need abstraction work to be in place, otherwise upgrading will be an big overhaul

packages/core/src/components/misc/CredentialCard10.tsx
```diff
@@ -143,23 +146,32 @@ const CredentialCard10: React.FC<CredentialCard10Props> = ({ credential, style =
...
-    }
-    bundleResolver.resolveAllBundles(params).then((bundle) => {
+    (async () => {
+      const identifiers = getCredentialIdentifiers(credential)
+      const attributes = buildFieldsFromAnonCredsCredential(credential)
+      const credName = await getCredentialName(
+        identifiers.credentialDefinitionId,
+        identifiers.schemaId,
+        agent
+      )
```
MacDeluca, line 156: suggestion: looks like a similar pattern is used a few times throughout this PR, might be a good place to write a custom hook ie: `useCredential` or similarly named where we can load and return these attributes => `{ identifiers, attributes, name }`

packages/core/src/components/misc/CredentialCard.tsx
```diff
@@ -79,16 +86,49 @@ const CredentialCard: React.FC<CredentialCardProps> = ({
     }
   }, [credential, brandingOverlay, resolveBundleForCredential])
 
+  // Resolve identifiers and name (supports webvh via agent) for anoncreds credentials
+  useEffect(() => {
+    let isMounted = true;
+    (async () => {
```
MacDeluca, line 92: nit: Avoid using immediately invoked anonymous functions, they can make the code harder to read. Consider a named function within the useEffect

```ts
useEffect(() => {
// Helpful comment about async operations
const asyncEffect = () => {
...
}
asyncEffect()

}, [])
```

packages/core/src/components/misc/CredentialCard10.tsx
```diff
@@ -143,23 +146,32 @@ const CredentialCard10: React.FC<CredentialCard10Props> = ({ credential, style =
...
-      identifiers: getCredentialIdentifiers(credential),
-      attributes: buildFieldsFromAnonCredsCredential(credential),
-      meta: {
-        credConnectionId: credential?.connectionId,
-        alias: credentialConnectionLabel,
-      },
-      language: i18n.language,
-    }
-    bundleResolver.resolveAllBundles(params).then((bundle) => {
+    (async () => {
```
MacDeluca, line 149: nit: Avoid using immediately invoked anonymous functions

packages/core/src/screens/CredentialDetails.tsx
```diff
@@ -161,24 +161,32 @@ const CredentialDetails: React.FC<CredentialDetailsProps> = ({ navigation, route
...
-    const params = {
-      identifiers: getCredentialIdentifiers(credential),
-      meta: {
-        alias: credentialConnectionLabel,
-        credConnectionId: credential.connectionId,
-      },
-      attributes: buildFieldsFromAnonCredsCredential(credential),
-      language: i18n.language,
-    }
+    (async () => {
```
MacDeluca, line 164: nit: Avoid using immediately invoked anonymous functions

bryce-mcmath:
Probably better to store schema info in metadata so we're hitting the ledger so frequently

Issue #1672, opened 2025-10-02
Cache schema when storing a credential


PatStLouis:
When bifold stores a credential and needs to display a label for a credential card or a presentation card, it currently derives this label from the (indy) schema ID. This creates issues with other DID method such as the latest work with WebVH. I want to suggest that bifold stores the schema as metadata of the credential when storing it, so it has access to the schema name in a more reliable way which is method agnostic.

### Acceptance Criteria

Bifold can display a credential label without having to resolve or parse a schemaId, based on a locally stored schema of said credential.

See PR #1667 for more context.

jleach:
@PatStLouis Was this completed already?

PatStLouis:
@jleach this is completed yes, closing

PR #1681, opened 2025-10-17
feat: add schema name caching and WebVH support

PatStLouis:
# Summary of Changes

This improves credential display by caching the schema name at offer time and using it when OCA overlays don't provide a specific name or only return generic names like 'Credential'. Also enables support for WebVH-based credentials.

- Add getSchemaName() utility to retrieve cached schema name from credential metadata
- Store schema name in AnonCredsCredentialMetadataKey when credentials are offered
- Use cached schema name as fallback when OCA overlay doesn't provide meaningful name
- Add WebVhAnonCredsRegistry and WebvhDidResolver for did:webvh credential support
- Update credential cards and details views to use cached schema name
- Optimize history logging to avoid unnecessary async calls when cached name exists
- Add tests for schema name retrieval and caching

# Testing Instructions

Issue and present a credential without OCA. The credential name should use the schema name value.

# Acceptance Criteria

Schema name is used for credential label when no oca bundle exists.

N/A

# Breaking change guide

If users where relying on cred def tags for display, this will no longer work.

# Related Issues

#1672

# Pull Request Checklist

- [X] All commits contain a DCO `Signed-off-by` line (we use the [DCO GitHub app](https://github.com/apps/dco) to enforce this)
- [X] If applicable, screenshots, gifs, or video are included for UI changes
- [X] If applicable, breaking changes are described above along with how to address them
- [X] If applicable, added [changeset(s)](https://github.com/changesets/changesets)
- [X] Added sufficient [tests](../packages/core/__tests__/) so that overall code coverage is not reduced

Pro Tip 🤓

### DIFF
--- a/.changeset/add-anoncreds-metadata-caching.md
+++ b/.changeset/add-anoncreds-metadata-caching.md
@@ -1,0 +1,6 @@
+---
+'@bifold/core': minor
+---
+
+Added AnonCreds metadata caching for schema name
+
--- a/.changeset/add-anoncreds-metadata-caching.md
+++ b/.changeset/add-anoncreds-metadata-caching.md
@@ -1,6 +1,6 @@
 ---
-'@bifold/core': minor
+'@bifold/core': patch
 ---
 
-Added AnonCreds metadata caching for schema name
+Added AnonCreds metadata caching for schema name and improved error handling
 
--- a/packages/core/__tests__/screens/__snapshots__/W3cProofRequest.test.tsx.snap
+++ b/packages/core/__tests__/screens/__snapshots__/W3cProofRequest.test.tsx.snap
@@ -886,7 +886,7 @@ exports[`displays a proof request screen ProofRequest Screen, W3C displays a pro
                   "threadId": "1",
                 },
                 "credId": "8eba4449-8a85-4954-b11c-e0590f39cbdb",
-                "credName": "latest",
+                "credName": "Credential",
                 "schemaId": "did:indy:bcovrin:test:TfuPA6whW681GfU6fj1e3k/anoncreds/v0/SCHEMA/Identity Schemaaf92af92-524d-41b6-bee8-00fb45e8eb6c/1.0.0",
               },
             ]
@@ -959,7 +959,7 @@ exports[`displays a proof request screen ProofRequest Screen, W3C displays a pro
                     style={
                       Array [
                         Object {
-                          "backgroundColor": "#5f0f7b",
+                          "backgroundColor": "#781e4b",
                           "borderRadius": 10,
                           "shadowColor": "#000",
                           "shadowOffset": Object {
@@ -1031,7 +1031,7 @@ exports[`displays a proof request screen ProofRequest Screen, W3C displays a pro
                         testID="com.ariesbifold:id/CredentialCard"
                       >
                         <View
-                          accessibilityLabel="Credentials.IssuedBy Unknown Contact,  Latest Credentials.Credential. email, test@email.com, time, 2022-02-11 20:00:18.180718"
+                          accessibilityLabel="Credentials.IssuedBy Unknown Contact,  Credential Credentials.Credential. email, test@email.com, time, 2022-02-11 20:00:18.180718"
                           accessible={true}
                           style={
                             Object {
@@ -1044,13 +1044,13 @@ exports[`displays a proof request screen ProofRequest Screen, W3C displays a pro
                             style={
                               Array [
                                 Object {
-                                  "backgroundColor": "#5f0f7b",
+                                  "backgroundColor": "#781e4b",
                                   "borderBottomLeftRadius": 10,
                                   "borderTopLeftRadius": 10,
                                   "width": 90,
                                 },
                                 Object {
-                                  "backgroundColor": "#5f0f7b",
+                                  "backgroundColor": "#781e4b",
                                   "overflow": "hidden",
                                 },
                               ]
@@ -1107,7 +1107,7 @@ exports[`displays a proof request screen ProofRequest Screen, W3C displays a pro
                               }
                               testID="com.ariesbifold:id/NoLogoText"
                             >
-                              L
+                              C
                             </Text>
                           </View>
                           <View
@@ -1154,7 +1154,7 @@ exports[`displays a proof request screen ProofRequest Screen, W3C displays a pro
                                 }
                                 testID="com.ariesbifold:id/CredentialName"
                               >
-                                Latest
+                                Credential
                               </Text>
                             </View>
                             <View
--- a/packages/core/src/components/misc/CredentialCard10.tsx
+++ b/packages/core/src/components/misc/CredentialCard10.tsx
@@ -26,7 +26,7 @@ import {
 import { formatTime, useCredentialConnectionLabel } from '../../utils/helpers'
 import { buildFieldsFromAnonCredsCredential } from '../../utils/oca'
 import { testIdWithKey } from '../../utils/testable'
-import { getSchemaName } from '../../utils/cred-def'
+import { getEffectiveCredentialName } from '../../utils/credential'
 
 import CardWatermark from './CardWatermark'
 
@@ -154,33 +154,22 @@ const CredentialCard10: React.FC<CredentialCard10Props> = ({ credential, style =
       language: i18n.language,
     }
     bundleResolver.resolveAllBundles(params).then((bundle) => {
-      // Use cached schema name as fallback if no specific overlay name is found
-      const cachedSchemaName = getSchemaName(credential)
-      // Check if OCA name is meaningful (not just "Credential" or empty)
-      const ocaName = bundle.metaOverlay?.name
-      const hasMeaningfulOcaName = ocaName && ocaName !== 'Credential' && ocaName.trim() !== ''
-
-      const effectiveName = hasMeaningfulOcaName ? ocaName : cachedSchemaName || ocaName
+      const effectiveName = getEffectiveCredentialName(credential, bundle.metaOverlay?.name)
 
       setOverlay((o) => ({
         ...o,
         ...bundle,
         brandingOverlay: bundle.brandingOverlay as LegacyBrandingOverlay,
+        // Apply effective name if different from OCA name
+        ...(effectiveName && effectiveName !== bundle.metaOverlay?.name && bundle.metaOverlay
+          ? {
+              metaOverlay: {
+                ...bundle.metaOverlay,
+                name: effectiveName,
+              } as any,
+            }
+          : {}),
       }))
-
-      // Update the overlay name after setting the overlay
-      if (effectiveName && effectiveName !== ocaName && bundle.metaOverlay) {
-        setOverlay((o) => {
-          const updatedOverlay = {
-            ...o,
-            metaOverlay: {
-              ...o.metaOverlay,
-              name: effectiveName,
-            } as any,
-          }
-          return updatedOverlay
-        })
-      }
     })
   }, [credential, credentialConnectionLabel, i18n.language, bundleResolver])
 
--- a/packages/core/src/components/misc/CredentialCard10.tsx
+++ b/packages/core/src/components/misc/CredentialCard10.tsx
@@ -26,6 +26,7 @@ import {
 import { formatTime, useCredentialConnectionLabel } from '../../utils/helpers'
 import { buildFieldsFromAnonCredsCredential } from '../../utils/oca'
 import { testIdWithKey } from '../../utils/testable'
+import { getSchemaName } from '../../utils/cred-def'
 
 import CardWatermark from './CardWatermark'
 
@@ -153,11 +154,33 @@ const CredentialCard10: React.FC<CredentialCard10Props> = ({ credential, style =
       language: i18n.language,
     }
     bundleResolver.resolveAllBundles(params).then((bundle) => {
+      // Use cached schema name as fallback if no specific overlay name is found
+      const cachedSchemaName = getSchemaName(credential)
+      // Check if OCA name is meaningful (not just "Credential" or empty)
+      const ocaName = bundle.metaOverlay?.name
+      const hasMeaningfulOcaName = ocaName && ocaName !== 'Credential' && ocaName.trim() !== ''
+
+      const effectiveName = hasMeaningfulOcaName ? ocaName : cachedSchemaName || ocaName
+
       setOverlay((o) => ({
         ...o,
         ...bundle,
         brandingOverlay: bundle.brandingOverlay as LegacyBrandingOverlay,
       }))
+
+      // Update the overlay name after setting the overlay
+      if (effectiveName && effectiveName !== ocaName && bundle.metaOverlay) {
+        setOverlay((o) => {
+          const updatedOverlay = {
+            ...o,
+            metaOverlay: {
+              ...o.metaOverlay,
+              name: effectiveName,
+            } as any,
+          }
+          return updatedOverlay
+        })
+      }
     })
   }, [credential, credentialConnectionLabel, i18n.language, bundleResolver])
 
--- a/packages/core/src/components/misc/CredentialCard10.tsx
+++ b/packages/core/src/components/misc/CredentialCard10.tsx
@@ -154,21 +154,15 @@ const CredentialCard10: React.FC<CredentialCard10Props> = ({ credential, style =
       language: i18n.language,
     }
     bundleResolver.resolveAllBundles(params).then((bundle) => {
-      const effectiveName = getEffectiveCredentialName(credential, bundle.metaOverlay?.name)
-
       setOverlay((o) => ({
         ...o,
         ...bundle,
         brandingOverlay: bundle.brandingOverlay as LegacyBrandingOverlay,
-        // Apply effective name if different from OCA name
-        ...(effectiveName && effectiveName !== bundle.metaOverlay?.name && bundle.metaOverlay
-          ? {
-              metaOverlay: {
-                ...bundle.metaOverlay,
-                name: effectiveName,
-              } as any,
-            }
-          : {}),
+        // Apply effective name
+        metaOverlay: {
+          ...bundle.metaOverlay,
+          name: getEffectiveCredentialName(credential, bundle.metaOverlay?.name),
+        } as any,
       }))
     })
   }, [credential, credentialConnectionLabel, i18n.language, bundleResolver])
--- a/packages/core/src/components/misc/CredentialCard11.tsx
+++ b/packages/core/src/components/misc/CredentialCard11.tsx
@@ -21,7 +21,7 @@ import {
 } from '../../utils/helpers'
 import { shadeIsLightOrDark, Shade } from '../../utils/luminance'
 import { testIdWithKey } from '../../utils/testable'
-import { getSchemaName } from '../../utils/cred-def'
+import { getEffectiveCredentialName } from '../../utils/credential'
 
 import CardWatermark from './CardWatermark'
 import CredentialActionFooter from './CredentialCard11ActionFooter'
@@ -237,32 +237,22 @@ const CredentialCard11: React.FC<CredentialCard11Props> = ({
         }
       }
 
-      // Use cached schema name as fallback if no specific overlay name is found
-      const cachedSchemaName = credential ? getSchemaName(credential) : undefined
-      const ocaName = bundle.metaOverlay?.name
-      const hasMeaningfulOcaName = ocaName && ocaName !== 'Credential' && ocaName.trim() !== ''
-
-      const effectiveName = hasMeaningfulOcaName ? ocaName : cachedSchemaName || ocaName
+      const effectiveName = credential ? getEffectiveCredentialName(credential, bundle.metaOverlay?.name) : undefined
 
       setOverlay((o) => ({
         ...o,
         ...bundle,
         brandingOverlay: bundle.brandingOverlay as BrandingOverlay,
+        // Apply effective name if different from OCA name
+        ...(effectiveName && effectiveName !== bundle.metaOverlay?.name && bundle.metaOverlay
+          ? {
+              metaOverlay: {
+                ...bundle.metaOverlay,
+                name: effectiveName,
+              } as any,
+            }
+          : {}),
       }))
-
-      // Update the overlay name after setting the overlay
-      if (effectiveName && effectiveName !== ocaName && bundle.metaOverlay) {
-        setOverlay((o) => {
-          const updatedOverlay = {
-            ...o,
-            metaOverlay: {
-              ...o.metaOverlay,
-              name: effectiveName,
-            } as any,
-          }
-          return updatedOverlay
-        })
-      }
     })
   }, [
     credential,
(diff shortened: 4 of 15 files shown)


jleach:
> This improves credential display by caching the schema name at offer time and using it when OCA overlays don't provide a specific name or only return generic names like 'Credential'. Also enables support for WebVH-based credentials.

How common is this? Could this be a custom use case to your current testing scenario? For examply why not "Le justificatif" in French or "Der Ausweis" in German. Seems like any issuer can use any variant they like - no?

packages/core/src/components/misc/CredentialCard10.tsx
```diff
@@ -153,11 +154,33 @@ const CredentialCard10: React.FC<CredentialCard10Props> = ({ credential, style =
       language: i18n.language,
     }
     bundleResolver.resolveAllBundles(params).then((bundle) => {
+      // Use cached schema name as fallback if no specific overlay name is found
+      const cachedSchemaName = getSchemaName(credential)
+      // Check if OCA name is meaningful (not just "Credential" or empty)
+      const ocaName = bundle.metaOverlay?.name
+      const hasMeaningfulOcaName = ocaName && ocaName !== 'Credential' && ocaName.trim() !== ''
```
jleach: Could this be a custom use case to your current testing scenario? For example why not "Le justificatif" in French or "Der Ausweis" in German rather than "Credential" in English? Where I'm going here is are we assuming the English word is always used here? I though OCA always had the built in concept of i8n - no?

packages/core/src/components/misc/CredentialCard11.tsx
```diff
@@ -235,11 +236,33 @@ const CredentialCard11: React.FC<CredentialCard11Props> = ({
           })
         }
       }
+
+      // Use cached schema name as fallback if no specific overlay name is found
+      const cachedSchemaName = credential ? getSchemaName(credential) : undefined
+      const ocaName = bundle.metaOverlay?.name
+      const hasMeaningfulOcaName = ocaName && ocaName !== 'Credential' && ocaName.trim() !== ''
```
jleach: Since "Credential" is used in a few places and may change depending on the native language of the wallet should this go in a constants file?

packages/core/src/components/misc/CredentialCard10.tsx
```diff
@@ -153,11 +154,33 @@ const CredentialCard10: React.FC<CredentialCard10Props> = ({ credential, style =
       language: i18n.language,
     }
     bundleResolver.resolveAllBundles(params).then((bundle) => {
+      // Use cached schema name as fallback if no specific overlay name is found
+      const cachedSchemaName = getSchemaName(credential)
```
jleach: Below you do `credential ? getSchemaName(credential) : undefined`. Should we do that here also for consistency?

packages/core/src/screens/CredentialDetails.tsx
```diff
@@ -172,14 +174,68 @@ const CredentialDetails: React.FC<CredentialDetailsProps> = ({ navigation, route
     }
 
     bundleResolver.resolveAllBundles(params).then((bundle) => {
+      // Use cached schema name as fallback if no specific overlay name is found
+      const cachedSchemaName = getSchemaName(credential)
```
jleach: In the spirit of keeping your code [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) should these three lines be moved into a helper function like `effectiveNameForCredential()`, similar to `getSchemaName` in cred-def.ts?

packages/core/src/screens/CredentialOfferAccept.tsx
```diff
@@ -84,8 +86,78 @@ const CredentialOfferAccept: React.FC<CredentialOfferAcceptProps> = ({ visible,
...
+                  logger.info('Restoring schema metadata for accepted credential', {
+                    credentialId: credential.id,
+                    schemaName: resolvedSchema.name,
+                    schemaId: offerData.schema_id,
+                  })
+
+                  // Store schema metadata
+                  const metadataToStore = {
+                    schemaId: offerData.schema_id,
+                    credentialDefinitionId: 'cred_def_id' in offerData ? offerData.cred_def_id : undefined,
```
jleach: Is this hard coded property `cred_def_id` specific to `webhv` or a test credential you're working with. Another way, will every webhv credential have this propert?

jleach (changes requested):

packages/core/src/components/misc/CredentialCard10.tsx
PatStLouis: 'Credential' seemed to be a hard value coded in my experience

packages/core/src/components/misc/CredentialCard11.tsx
PatStLouis: Let me review where it was already used in the code, it was already a fall back value. I like the idea of making it a variable DefaultCredentialNameValue or something to make it clearer

packages/core/src/components/misc/CredentialCard10.tsx
PatStLouis: I will look into it

packages/core/src/screens/CredentialDetails.tsx
PatStLouis: I think so, just had a chat with Bryce and we came to the same conclusion. I will rebase with the webvh pr that was merged and clean this up

packages/core/src/screens/CredentialOfferAccept.tsx
PatStLouis: This comes from credo and is already a default value, not webvh sepcific:
https://github.com/openwallet-foundation/bifold-wallet/blob/165cb5388d2ec57ecdfd4baf0812cc5317101c3f/packages/core/src/screens/CredentialOffer.tsx#L130

packages/core/src/screens/CredentialDetails.tsx
PatStLouis: I've implemented an getEffectiveCredentialName function

packages/core/src/components/misc/CredentialCard10.tsx
PatStLouis: implemented a shared function

packages/core/src/components/misc/CredentialCard11.tsx
PatStLouis: added a global variable

.changeset/add-anoncreds-metadata-caching.md
bryce-mcmath, line 1: Can we please make this a patch changeset instead of a minor one? We don't follow semantic versioning as new features and breaking changes happen almost every week so the version would balloon to a crazy high number so quickly.

packages/core/__tests__/screens/__snapshots__/W3cProofRequest.test.tsx.snap
```diff
@@ -1044,13 +1044,13 @@ exports[`displays a proof request screen ProofRequest Screen, W3C displays a pro
                             style={
                               Array [
                                 Object {
-                                  "backgroundColor": "#5f0f7b",
+                                  "backgroundColor": "#781e4b",
```
bryce-mcmath, line 1047: Q: What is causing this background colour to change?

packages/core/src/components/views/CredentialDetailPrimaryHeader.tsx
```diff
@@ -30,6 +31,10 @@ const CredentialDetailPrimaryHeader: React.FC<CredentialDetailPrimaryHeaderProps
   const { ColorPalette } = useTheme()
   const { width, height } = useWindowDimensions()
   const isBranding11 = brandingOverlayType === BrandingOverlayType.Branding11
+
+  // Use cached schema name as fallback if no specific overlay name is found
```
bryce-mcmath: nit: I don't think this comment is needed here, probably suffices to have a similar comment in the actual getEffectiveCredentialName function itself

packages/core/src/screens/CredentialDetails.tsx
```diff
@@ -176,10 +176,27 @@ const CredentialDetails: React.FC<CredentialDetailsProps> = ({ navigation, route
...
   }, [credential, credentialConnectionLabel, bundleResolver, i18n.language])
 
+  // Ensure credential has all required metadata
+  useEffect(() => {
+    if (!credential || !isValidAnonCredsCredential(credential) || !agent) {
+      return
+    }
+
+    const restoreMetadata = async () => {
+      await ensureCredentialMetadata(credential, agent)
```
bryce-mcmath: If it's possible for this async function  to throw an error, it should be wrapped in a try catch block

packages/core/src/screens/CredentialOffer.tsx
```diff
@@ -128,9 +126,10 @@ const CredentialOffer: React.FC<CredentialOfferProps> = ({ navigation, credentia
       const offerData = offer?.anoncreds ?? offer?.indy
 
       if (offerData) {
-        credential.metadata.add(AnonCredsCredentialMetadataKey, {
-          schemaId: offerData.schema_id,
-          credentialDefinitionId: offerData.cred_def_id,
+        // Ensure credential has all required metadata
```
bryce-mcmath: nit: This comment is kind of a restatement of the function name and is not needed imo

packages/core/src/screens/CredentialOffer.tsx
```diff
@@ -128,9 +126,10 @@ const CredentialOffer: React.FC<CredentialOfferProps> = ({ navigation, credentia
...
+        await ensureCredentialMetadata(credential, agent, {
+          schema_id: offerData.schema_id,
+          cred_def_id: offerData.cred_def_id,
         })
```
bryce-mcmath: If this awaited function can throw, it should probably be wrapped in a try catch block

packages/core/src/screens/CredentialOfferAccept.tsx
```diff
@@ -84,8 +86,16 @@ const CredentialOfferAccept: React.FC<CredentialOfferAcceptProps> = ({ visible,
     if (credential.state === CredentialState.CredentialReceived || credential.state === CredentialState.Done) {
       timer && clearTimeout(timer)
       setCredentialDeliveryStatus(DeliveryStatus.Completed)
+
+      // Ensure credential has all required metadata (function checks what's missing)
```
bryce-mcmath: nit: same thing as above, comment can be removed

packages/core/src/screens/CredentialOfferAccept.tsx
```diff
@@ -84,8 +86,16 @@ const CredentialOfferAccept: React.FC<CredentialOfferAcceptProps> = ({ visible,
...
+      const restoreMetadata = async () => {
+        if (agent) {
+          await ensureCredentialMetadata(credential, agent)
```
bryce-mcmath: Same thing as above, wrap in try catch if it can throw

packages/core/src/utils/cred-def.ts
```diff
@@ -86,8 +93,35 @@ function credentialDefinition(credential: CredentialRecord): string | undefined
...
+  const metadata = credential.metadata.get(AnonCredsCredentialMetadataKey)
+  const credDefTag = metadata?.credDefTag
+  return credDefTag
+}
+
 export async function parsedCredDefNameFromCredential(credential: CredentialRecord, agent?: Agent): Promise<string> {
-  return getCredentialName(credentialDefinition(credential), credentialSchema(credential), agent)
+  // Ensure metadata is cached if agent is provided
+  if (agent) {
+    await ensureCredentialMetadata(credential, agent)
```
bryce-mcmath: Same as above; try catch

packages/core/src/utils/credential.ts
```diff
@@ -35,3 +37,143 @@ export const getCredentialIdentifiers = (credential: CredentialExchangeRecord) =
...
+  let credDefId = offerData?.cred_def_id || existingMetadata?.credentialDefinitionId
+
+  // If we still don't have IDs, try to get them from format data
+  if (!schemaId || !credDefId) {
+    try {
+      const { offer } = await agent.credentials.getFormatData(credential.id)
+      const formatOfferData = offer?.anoncreds ?? offer?.indy
+      if (formatOfferData && typeof formatOfferData === 'object') {
+        schemaId = schemaId || ('schema_id' in formatOfferData ? formatOfferData.schema_id : undefined)
+        credDefId = credDefId || ('cred_def_id' in formatOfferData ? formatOfferData.cred_def_id : undefined)
```
bryce-mcmath: Would this be simpler?

```suggestion
        schemaId = schemaId || formatOfferData?.schema_id
        credDefId = credDefId || formatOfferData?.cred_def_id
```

packages/core/src/utils/credential.ts
```diff
@@ -35,3 +37,143 @@ export const getCredentialIdentifiers = (credential: CredentialExchangeRecord) =
...
+export function getEffectiveCredentialName(credential: CredentialExchangeRecord, ocaName?: string): string {
+  // Helper function to validate a name
+  const isValidName = (name: string | undefined): boolean => {
+    return !!(
+      name &&
+      name !== defaultCredDefTag &&
+      name !== fallbackDefaultCredentialNameValue &&
+      name.trim() !== ''
+    )
+  }
```
bryce-mcmath: This function can be moved out of the parent function so it doesn't need to be redeclared every time the parent gets called. Would also make it easier to test it independently

bryce-mcmath (changes requested):
Small nits, good work so far

packages/core/__tests__/screens/__snapshots__/W3cProofRequest.test.tsx.snap
PatStLouis: > 7 Months Ago (April 2025):
> Commit b2987d7b: Filipe updated the OCA Branding code
> The actual running code changed to use #781e4b
> BUT the test snapshot file wasn't fully updated - it still had some old #5f0f7b values

> Yesterday (Your Commit ec60e0d0):
> You made changes to credential naming logic
> You ran the tests
> Jest noticed: "Hey, the snapshot says #5f0f7b but the actual output is #781e4b"
> Jest updated the snapshot to match what the code actually produces

PatStLouis: does this make sense to you?

packages/core/src/screens/CredentialDetails.tsx
PatStLouis: done

packages/core/src/screens/CredentialOffer.tsx
PatStLouis: comment removed

packages/core/src/screens/CredentialOfferAccept.tsx
PatStLouis: comment removed

packages/core/src/utils/cred-def.ts
PatStLouis: done

packages/core/src/utils/credential.ts
PatStLouis: had to do slightly different:
```
schemaId = schemaId || (formatOfferData as any).schema_id
credDefId = credDefId || (formatOfferData as any).cred_def_id
```

packages/core/src/utils/credential.ts
PatStLouis: done, test added

packages/core/src/screens/CredentialOfferAccept.tsx
PatStLouis: done

packages/core/src/screens/CredentialOffer.tsx
PatStLouis: done

packages/core/__tests__/screens/__snapshots__/W3cProofRequest.test.tsx.snap
bryce-mcmath: Not really just because snapshot changes are detected whenever the tests run and tests run as part of CICD so it would definitely be picked up and fail CICD in the last seven months.

Thinking more about it, I'm guessing the credential name is used to generate a random hex colour, and since it changed, so did the colour.

bryce-mcmath: Yes, the random colour is generated from a hash of the credential definition ID, so it makes sense that it would change since the cred def tag was changed. All good

packages/core/src/components/views/CredentialDetailPrimaryHeader.tsx
PatStLouis: removed

packages/core/src/utils/credential.ts
bryce-mcmath: `as any` should be avoided as much as possible

if it's possible for formatData to be undefined or null or a non-object then we should still use optional chaining. If you still have to cast with `as` try to use the correct type, not `any`

.changeset/add-anoncreds-metadata-caching.md
PatStLouis: I just made this into a patch

bryce-mcmath (dismissed):

packages/core/src/utils/credential.ts
PatStLouis: I imported the AnonCredsCredentialOffer type

packages/core/src/screens/CredentialDetails.tsx
```diff
@@ -176,10 +176,32 @@ const CredentialDetails: React.FC<CredentialDetailsProps> = ({ navigation, route
...
+    if (!credential || !isValidAnonCredsCredential(credential) || !agent) {
+      return
+    }
+
+    const restoreMetadata = async () => {
+      try {
+        await ensureCredentialMetadata(credential, agent)
+      } catch (error) {
+        // If metadata restoration fails, we'll fall back to default credential name
+        agent?.config.logger?.warn('Failed to restore credential metadata', { error: error as Error })
```
jleach: By convention we either pass in logger or get it from dependency injection.

```TypeScript
  const [logger] = useServices([TOKENS.UTIL_LOGGER])
```

packages/core/src/screens/CredentialOfferAccept.tsx
```diff
@@ -84,8 +86,19 @@ const CredentialOfferAccept: React.FC<CredentialOfferAcceptProps> = ({ visible,
     if (credential.state === CredentialState.CredentialReceived || credential.state === CredentialState.Done) {
       timer && clearTimeout(timer)
       setCredentialDeliveryStatus(DeliveryStatus.Completed)
+
+      const restoreMetadata = async () => {
+        if (agent) {
+          try {
+            await ensureCredentialMetadata(credential, agent)
+          } catch (error) {
+            agent.config.logger?.warn('Failed to restore credential metadata', { error: error as Error })
```
jleach: `logger.warn()` no?

packages/core/src/utils/cred-def.ts
```diff
@@ -31,7 +38,7 @@ export async function getCredentialName(credDefId?: string, schemaId?: string, a
 }
 
 async function parseWebVHCredDefId(credDefId?: string, schemaId?: string, agent?: Agent): Promise<string> {
```
jleach: Maybe this fn just needed to pass in logger rather than agent?

packages/core/src/utils/cred-def.ts
```diff
@@ -86,8 +93,40 @@ function credentialDefinition(credential: CredentialRecord): string | undefined
...
+
 export async function parsedCredDefNameFromCredential(credential: CredentialRecord, agent?: Agent): Promise<string> {
-  return getCredentialName(credentialDefinition(credential), credentialSchema(credential), agent)
+  // Ensure metadata is cached if agent is provided
+  if (agent) {
+    try {
+      await ensureCredentialMetadata(credential, agent)
+    } catch (error) {
+      // If metadata restoration fails, we'll fall back to parsing IDs or default name
+      agent?.config.logger?.warn('Failed to restore credential metadata in parsedCredDefNameFromCredential', { error: error as Error })
```
jleach: I think passing in logger would be better. You can only write log messages if agent, which is optional, is passed in. Also, if this is a `catch` maybe you mean .error() rather than .warn()?

packages/core/src/utils/credential.ts
```diff
@@ -35,3 +37,149 @@ export const getCredentialIdentifiers = (credential: CredentialExchangeRecord) =
...
+  if (!schemaId || !credDefId) {
+    try {
+      const { offer } = await agent.credentials.getFormatData(credential.id)
+      const formatOfferData = (offer?.anoncreds ?? offer?.indy) as AnonCredsCredentialOffer | undefined
+      if (formatOfferData) {
+        schemaId = schemaId || formatOfferData?.schema_id
+        credDefId = credDefId || formatOfferData?.cred_def_id
+      }
+    } catch (error) {
+      agent.config.logger?.warn('Failed to get format data', { error: error as Error })
```
jleach: `logger.error()`? This seems like an error condition.

packages/core/src/utils/helpers.ts
```diff
@@ -338,11 +339,46 @@ const credNameFromRestriction = async (
     cred_def_id = (query?.cred_def_id as string) ?? cred_def_id
     schema_id = (query?.schema_id as string) ?? schema_id
   })
-  if (schema_name && (schema_name.toLowerCase() !== 'default' || schema_name.toLowerCase() !== 'credential')) {
+
+  // Priority 1: Provided restriction schema name (from schema_name in restriction if meaningful)
```
jleach, line 375: Not critical for this PR but this fn is getting a bit long and hard to test. Writing 3 smaller functions then this one to just wrap 'em and return might make it easier to test:
```
getRestrictionSchemaName()
resolveCredentialDefenitionTagAsName()
resolveSchemaName()
```

jleach (changes requested):
- Some time ago Clecio asked us to rely less on `agent?.config.logger?` and use he one from dependency injection or pass it in to the function as an optional param.
- There is one really long fn. When they get overly long they're harder to test and maintain. Consider breaking it down into a few smaller functions and the main one just wraps them. Ref: https://en.wikipedia.org/wiki/Single-responsibility_principle
- Sonar noted that your imports need to be cleaned up. There are quite a few multiple imports:
['../../utils/credential' imported multiple times.](https://sonarcloud.io/project/issues?pullRequest=1681&open=AZoDndi7yhGybQZXtr-5&id=openwallet-foundation_bifold-wallet)

packages/core/src/utils/helpers.ts
PatStLouis: I created 2 functions for the resolving operations. Since the schema name if just 1 conditional check, I say its fine to leave it

packages/core/src/utils/cred-def.ts
PatStLouis: I passed in the logger, but the agent is still needed

PatStLouis: I've updated the code, the agent isn't needed anymore. I changed the webvh tests to use the new resolve function.

packages/core/src/utils/helpers.ts
PatStLouis: I updated the webvh tests to use these 2 new functions

PR #1711, opened 2025-11-16
feat: app store receipts

bryce-mcmath:
# Summary of Changes

Adds a simple app store receipt checking method to our attestation lib

I also exported the MockLogger to make testing output in downstream projects less cluttered

# Testing Instructions

N/A - needs to be from an official app store release

# Acceptance Criteria

When used in a downstream project from the app store, an app store receipt can be generated
 
N/A

# Breaking change guide

N/A

# Related Issues

- bcgov/bc-wallet-mobile#2752

# Pull Request Checklist

Pro Tip 🤓


packages/react-native-attestation/src/index.ts
```diff
@@ -93,3 +93,12 @@ export const isPlayIntegrityAvailable = async (): Promise<boolean> => {
...
   return Attestation.isPlayIntegrityAvailable();
 };
+
+export const getAppStoreReceipt = async (): Promise<string | undefined> => {
+  if (Platform.OS !== 'ios') {
+    throw new Error('getAppStoreReceipt is only available on iOS');
+  }
+
+  const result = await Attestation.getAppStoreReceipt();
+  return result ?? undefined;
```
MacDeluca: question: when does getAppStoreReceipt return nullish?

I guess what i'm asking is if this should be throwing an error or if we should return a null value instead.

packages/core/src/testing/MockLogger.ts
```diff
@@ -0,0 +1,16 @@
+import { AbstractBifoldLogger } from '../services/AbstractBifoldLogger'
+
+export class MockLogger extends AbstractBifoldLogger {
```
jleach, line 3: I thought these generally went into `__mock__` so that when you import
`import { appLogger, createAppLogger } from '@utils/logger'`

You get the mock object not the real one?

jleach (changes requested):

al-rosenthal (dismissed):

packages/core/src/testing/MockLogger.ts
bryce-mcmath: @jleach In order to export it I had to put it here. `__mocks__` and` __tests__` get tossed in the release process

packages/react-native-attestation/src/index.ts
bryce-mcmath: when it's a non-app store local build. updated to just return null. I don't think it should throw if it's just null, we can let the project using the package decide if they want to throw in that case

MacDeluca (approved):

packages/core/src/testing/MockLogger.ts
jleach: Wondering if we can add it to the npm package at some other time.

jleach (approved):

PR #1806, opened 2026-03-19
feat: expo app integrity initialization

LippaC-OPS:
# Summary of Changes

Added initialization setup function for Expo App Integrity package for app attestation. 

# Testing Instructions

Install Expo App Integrity in sample project, add mock functionality for getChallenge and getJWT functions, enable app attestation in config, start the app and confirm that the functions are called and app starts correctly after authentication

# Acceptance Criteria

Expo App Integrity setup correctly, without causing issues when disabled 

N/A

# Breaking change guide

This will require apps using Bifold to add Expo App Integrity to their application, even if they are not using app attestation here

# Related Issues

# Pull Request Checklist

Pro Tip 🤓


packages/core/src/hooks/attestation.ts
```diff
@@ -0,0 +1,146 @@
...
+            return res
+          })
+          .catch((err: any) => {
+            throw new Error(err)
+          })
+      }
+      else if (Platform.OS === 'android') {
+        prepareIntegrityTokenProviderAsync(attestation?.cloudProjectNumber)
+          .then(async (res) => {
+            await PersistentStorage.storeValueForKey(LocalStorageKeys.Attestation, '')
```
MosCD3: you can `agent.genericRecords` to save and retrieve data which has the benefit of biometrics lock

packages/core/src/types/config.ts
```diff
@@ -79,3 +80,20 @@ export interface HistoryEventsLoggerConfig {
...
+    challenge: string
+    platform: "ios" | "android" | "windows" | "macos" | "web"
+}
+
+export interface AttestationConfig {
+  applicationID: string
+  getAttestationChallenge: () => Promise<string>
+  getAttestationJWT: (params: getAttestationJWTParams) => Promise<any>
+  cloudProjectNumber: string
+  enableAttestation: boolean
```
MosCD3: enableAttestation to be the only config available

packages/core/src/container-api.ts
```diff
@@ -239,6 +260,7 @@ export type TokenMapping = {
   [TOKENS.UTIL_AGENT_BRIDGE]: AgentBridge
   [TOKENS.UTIL_REFRESH_ORCHESTRATOR]: IRefreshOrchestrator
   [TOKENS.FN_PIN_HASH_ALGORITHM]: FN_PIN_HASH_ALGORITHM
+  [TOKENS.ATTESTATION_CONFIG]: AttestationConfig
```
MosCD3: add an attestation getter token which will pass attestation object and get back attestation JWT

MosCD3 (dismissed):

packages/core/src/hooks/attestation.ts
```diff
@@ -0,0 +1,109 @@
...
+      if (Platform.OS === 'ios') {
+
+        const keyId = await generateKeyAsync()
+        const attestationResult = await generateAssertionAsync(keyId, challenge)
+        const attestationJWT = await getAttesatationJWT(attestationResult, challenge)
+        await storeAttestationJWT(attestationJWT, keyId)
+
+      } else if (Platform.OS === 'android') {
+
+        await prepareIntegrityTokenProviderAsync(attestation?.cloudProjectNumber)
```
timbl-ont: This creates a point in time integrity verdict. It is only valid for a day at best. Replace this with 

- AppIntegrity.generateHardwareAttestedKeyAsync(keyAlias, challenge)
- AppIntegrity.getAttestationCertificateChainAsync(keyAlias)

This is equivalent to what IOS is doing and creates an attestation that will live until the app is updated.

packages/core/src/types/config.ts
LippaC-OPS: done

packages/core/src/container-api.ts
LippaC-OPS: added

packages/core/src/hooks/attestation.ts
LippaC-OPS: replaced with those

MosCD3 (commented):
FN_ATTESTATION_GET_CHALLENGE
FN_ATTESTATION_GET_JWT 

Must have a default registration in the container implementation file in the core package. 

See the attached crash

Image

packages/core/src/onboarding.ts
```diff
@@ -50,8 +50,8 @@ export const isAttemptLockoutComplete = (servedPenalty: boolean | undefined): On
   return { name: Screens.AttemptLockout, completed: servedPenalty !== false }
 }
 
-export const isAgentInitializationComplete = (agent: Agent | null): OnboardingTask => {
-  return { name: Screens.Splash, completed: !!agent }
+export const isAgentInitializationComplete = (agent: Agent | null, isAttestationComplete: boolean): OnboardingTask => {
```
MosCD3, line 53: This will block completion if "enableAttestation" is false for non attestation enabled apps!! 
maybe use that instead 
`!!agent && (!config.enableAttestation || isAttestationComplete)`

LippaC-OPS:
@MosCD3 Ok will push a fix for that

samples/app/ios/PrivacyInfo.xcprivacy
```diff
@@ -6,22 +6,22 @@
 	<array>
 		<dict>
 			<key>NSPrivacyAccessedAPIType</key>
-			<string>NSPrivacyAccessedAPICategoryUserDefaults</string>
+			<string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
```
MosCD3, line 9: is that required in this PR?

packages/core/package.json
```diff
@@ -57,6 +57,7 @@
     "@credo-ts/question-answer": "0.6.1",
     "@credo-ts/react-native": "0.6.1",
     "@credo-ts/webvh": "0.6.1",
+    "@expo/app-integrity": "^55.0.9",
```
MosCD3, line 60: can you also add this to 'peerDependencies' please
Since downstream apps now must get a warning if this dependency is not installed.

packages/core/src/hooks/attestation.ts
```diff
@@ -0,0 +1,98 @@
...
+        })
+        await PersistentStorage.storeValueForKey(LocalStorageKeys.AttestationConfigured, true)
+        await PersistentStorage.storeValueForKey(LocalStorageKeys.AttestationKey, keyID)
+        dispatch({ type: DispatchAction.SET_ATTESTATION_COMPLETED, payload: [true] })
+      })
+    } catch (err: any) {
+      logger.error(err?.message ?? 'Error initializing attestation')
+      throw new Error('Error storing attestation result')
+    }
+  }, [agentBridge, dispatch, logger])
```
MosCD3, line 51: this function has 2 issues. 
1- the try/catch here will not fire, even if there is an error in attestation, because the agentBridge.onReady is synchronous and any error inside it will not bubble up. So moving the try/catch inside the listener will catch the failure


2. `onReady` is better suited for synchronous “agent is available” notifications, like the refresh flow:

```
  bridge.onReady((agent) => {
    this.agent = agent
    this.logger.info('🪝 [RefreshOrchestrator] Agent ready')
    if (this.opts.autoStart && this.opts.intervalMs) this.start()
  }, true)
```
In that case the callback only updates local state, logs, and optionally starts the orchestrator. There is no async work that the caller needs to wait for.
For attestation storage, the callback performs several awaited operations:

`  await agent.genericRecords.save(...)`
..etc

`AgentBridge.onReady` does not await the listener’s returned promise, so errors inside that async callback won’t be caught by the outer try/catch, and the caller may continue before the JWT is actually saved. 
For this case, `await agentBridge.waitForReady()` is a better fit because it gives us the agent first, then we can run the persistence steps in a normal awaited try/catch flow.

MosCD3:
I’d also add a startup/onboarding test for wallets that do not enable attestation. It's important after you pushed your fix for default implementation

MosCD3:
can you rebase and resolve conflicts, you probably will want to push your version of Podlock

bryce-mcmath (dismissed):
It looks like this PR is a bit opinionated. If I'm reading it correctly it will force all downstream apps to install expo-app-integrity even if they don't intend to use it at all. It would probably be best to keep it injectable and optional

timbl-ont:
@bryce-mcmath We have also debated this internally. 

My take is that app integrity is part of a core standard - openid4vci. The wallet supports core standards of various kinds being a multi-protocol wallet. Not every derived wallet uses all the core standards but they are still included as libraries - for example indy-vdr which we don't use but it is still need to be included in builds

The library just exposes the OS level functions to initiate key attestation. It does the same thing more or less as the attestation package that is already in the bifold repo (plus Android key attestation) and would be similar to a library that exposes something like the keystore (react-native-keychain and expo-secure-store). 

The code to actually initiate an attestation is injectable. This backend piece is opinionated and will depend on each particular wallet implementation.

If we make the core library injectable than any wallet implementer who wants to implement openid4vci will need to figure out a custom solution for part of the core standard in the wallet.

The choice of the expo library aligns with our approach of moving to core expo functionality.

All of this does speak to looking at a more modular approach to including various specs, but that is future.

timbl-ont:
To clarify - the only thing that is opinionated with respect to bifold core in this PR is the choice of library.

- The implementation of key attestation is implemented at the IOS and Android level. Like the keychain the OS level API's are fixed.
- The use of key attestation is defined in the openid4vci specification.

samples/app/container-imp.tsx
```diff
@@ -56,6 +55,49 @@ export class AppContainer implements Container {
         throw new Error(`Error generating hash for PIN ${String((error as Error)?.message ?? error)}`)
       }
     })
+    */
```
al-rosenthal, line 58: if you think this should be left in, can you add a comment explaining what it is and why it's commented out

packages/core/src/hooks/attestation.ts
```diff
@@ -0,0 +1,98 @@
...
+        if (!isDeviceAttestationSupported) throw new Error('iOS device not supported')
+
+        const keyID = await generateKeyAsync()
+        const attestationResult = await withRetry(attestKeyAsync, [keyID, challenge])
+        const attestationJWT = await getAttestationJWT(attestationResult, challenge, keyID)
+        await storeAttestationJWT(attestationJWT, keyID)
+
+      } else if (Platform.OS === 'android') {
+
+        const keyID = 'walletAttestationKey'
```
al-rosenthal: should this be `generateKeyAsync()`?

al-rosenthal:
Attestation is already in it’s own module so I think it would make sense to decouple this from /core and move it into the /react-native-attestation module - keep the new definitions and flag and move the logic over. This way downstream uses of the wallet can opt in and use the provided attestation module or roll their own if they’d like. 

What do you guys think @bryce-mcmath @timbl-ont

timbl-ont:
@al-rosenthal To be clear attestation as a working standards compliant function is not in the attestation module (package). The attestation module is an implementation of OS specific attestation functions bridged to react-native - i.e. a library that does the same thing as expo app integrity but missing Android key attestation. This package pre-dated the availability of the expo library and bifold support for expo. 

If this PR is moved to the current bifold attestation package it would be one line to include the expo library and that is it. The actual logic that allows a standards based use of attestation would still need to be in core or a derived wallet - the logic in this PR. That gets back to my original argument above that the logic is in support of a standards compliant oid4vci specification implementation and as such belongs in core.

I think that with the release of the expo app integrity library the bifold attestation module should perhaps be deprecated? It has not been updated in three years and is missing functionality included in the expo library.

packages/core/src/hooks/attestation.ts
LippaC-OPS: `generateKeyAsync()` is the iOS specific function for Android its `generateHardwareAttestedKeyAsync()` which then the key is accessed by the keyID labelled in the line above. 

What I'm not sure about here is if the keyID should be randomly generated instead of hardcoded like this?

samples/app/container-imp.tsx
LippaC-OPS: It might be best here to remove it, since the types describe what the functions should return to the core to provide whats needed for attestation

packages/core/src/hooks/attestation.ts
LippaC-OPS: I ended up changing the keyID to be a key generated with the react-native-uuid library that was already included in the project

MosCD3 (approved):

